PR #1, opened 2017-01-27
Add mil_perception package and start mil_vision_lib

kev-the-dev:
- Create package mil_perception for containing well tested perception tools and nodes not specific to one MIL project
- Create mil_vision_lib in mil_perception, a C++ library with non-project specific vision tools
- Add a function to mil_vision_lib for parsing opencv 3D points from a ROS param
- Add a function to mil_vision_lib for displaying an image of (0,0,0) centered 3D points, useful for verifying you have modelled an object correctly for use in solvePnP


mil_perception/include/mil_vision_lib/cv_tools.hpp
```diff
@@ -0,0 +1,26 @@
...
+    /** Retrieves a paramater of a list of list of doubles and puts them into a vector of opencv 3D points
+     * @param topic Ros topic in the global namespace to load points from
+     * @param points A vector of cv::Point3f to put points in from the topic
+     * @return True if points loaded successfully, False if an error occurred (details will be in ros error output)
+     */
+    bool get3DPointsFromParam(std::string topic, std::vector<cv::Point3f>& points);
+    /// @param value A ROS XML value for parsing points from an already resolved param
+    bool get3DPointsFromParam(XmlRpc::XmlRpcValue& value, std::vector<cv::Point3f>& points);
+    /// @param nh Ros nodehandle to retrieve topic from
+    /// @param topic Ros topic in nh's namespace to load points from
```
DSsoto: This is a param not a topic, you should change this everywhere else you're calling a param a topic.

mil_perception/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(mil_perception)
+
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 ")
+
+find_package(catkin
```
DSsoto, line 6: You're going to want eigen3 here.

mil_perception/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+add_library(mil_vision_lib
+   src/mil_vision_lib/cv_tools.cpp
+)
+
+target_link_libraries(mil_vision_lib
+   ${catkin_LIBRARIES}
+   ${OpenCV_LIBS}
+)
+
+install(TARGETS mil_vision_lib
```
DSsoto: This isn't really necessary, catkin will still do this without this. We're not going to change the default install location.

mil_perception/include/mil_vision_lib/cv_tools.hpp
```diff
@@ -0,0 +1,26 @@
+#include <string>
+#include <vector>
+#include <opencv2/opencv.hpp>
```
DSsoto, line 3: These kind of libraries usually take the longest to compile since they use so many things that are pretty complex on their own. I'd recommend not using this convenience include but instead including the things that are actually used.

mil_perception/src/mil_vision_lib/cv_tools.cpp
```diff
@@ -0,0 +1,70 @@
+#include <mil_vision_lib/cv_tools.hpp>
+
+bool mil_vision::get3DPointsFromParam(std::string topic, std::vector<cv::Point3f>& points)
+{
+    XmlRpc::XmlRpcValue list;
```
DSsoto: It would be better to use a std::list here. It's the more standard way to do it now and it's a much more simple implementation.

mil_perception/src/mil_vision_lib/cv_tools.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (!ros::param::get(topic, list)) return false;
+    return mil_vision::get3DPointsFromParam(list, points);
+}
+bool mil_vision::get3DPointsFromParam(ros::NodeHandle& nh, std::string topic, std::vector<cv::Point3f>& points)
+{
+    XmlRpc::XmlRpcValue list;
+    if (!nh.getParam(topic, list)) return false;
+    return mil_vision::get3DPointsFromParam(list, points);   
+}
+bool mil_vision::get3DPointsFromParam(XmlRpc::XmlRpcValue& list, std::vector<cv::Point3f>& points)
```
DSsoto: I'd say that this is an abuse of overloading since the functions are semantically different. This is not really recursive although the naming could lead one to blelieve so at a glance.

DSsoto (changes requested):
Requesting a couple of changes.

mil_dev_tools/mil_dev_tools/msg_helpers.py
```diff
@@ -0,0 +1,195 @@
+import numpy as np
```
mattlangford, line 1: Please credit this to jake

mil_dev_tools/mil_dev_tools/text_effects.py
```diff
@@ -0,0 +1,146 @@
+class Colors():
+    red = '\033[31m'
```
mattlangford, line 2: Please credit this to me

mil_dev_tools/package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<package>
+  <name>mil_dev_tools</name>
+  <version>0.0.0</version>
+  <description>General scripts, libraries, and tools for MIL projects</description>
+
+  <maintainer email="kma1660@gmail.com">Kevin Allen</maintainer>
```
mattlangford, line 7: I think David, you, and I should be here

mil_dev_tools/mil_dev_tools/msg_helpers.py
mattlangford: And what about all the other tools?

mattlangford (changes requested):
Gotta give that credit where it's due

Also, look to use python for little tools like this. Python would be able to do the same thing in far fewer lines due to the automatic rosparam type deduction.

sentree:
Any tools that are being ported to software-common need to maintain their history! I have not looked into how to properly do this yet, but @forrestv did it for the installer repository with install.sh and I'm sure some light googling will turn something up.

kev-the-dev:
Addressed David's C++ stuff and did my best to preserve history of the tools while pruning the rest. It's going to be a bit of a pain to do this for all the other thing's we're importing from navigator/sub8, but I agree it's worth it.

mil_dev_tools/mil_dev_tools/move_helper.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
mattlangford: Can you remove this

mattlangford (commented):
Nice job, thanks for putting in the work
I had one request (to remove a file) and then I'm good.

sentree:
This should not be merged as it is. If  we bring tools over one by one, cherry picking files out of commits along the way, we will almost certainly end up with duplicates. #2 should be used to track the packages that we need to move and we should decide on a deadline date at which point they will all be moved over together. This will keep the history clean and prevent issues down the road. I propose we allow a week to post tools that need to be moved, which would mean that the final move will happen on 2/9/16 at 9p.

DSsoto:
This has been delayed a little but I agree with @whispercoros that it should be done all at once. I'll try to move perception and sonar stuff sometime soon.

@ironmig  could you address my comment regarding includes for cv_tools?

kev-the-dev:
So we are going to migrate everything at once so this PR is going to be closed. The branch still contains the vision lib stuff and the necessary refactoring of the tools to work with the new package, so use it for diffs

PR #3, opened 2017-01-31
ML: Adds a labeler script with various tools

mattlangford:
This is the start of the machine learning framework for use on the sub.
This tool is an easy to use tool for segmenting or labeling images. Right now there are three tools and two so-called image providers (more to come of both!)

**Tools:**
- Paint brush tool: paints the mask like a brush
- Polygon tool: Lets you draw polygons
- Move tool: Lets you move the mask around

**Image Providers:**
- Random color: mostly used for testing
- Webcam: gives you frame by frame access to your webcam

The goal with this is to be an easy to improve framework for segmenting images that will soon be used to train a learned segmenter. The requirements for adding new tools or image providers are fairly straight forward and just require simple adaptions to the current implementation.

P.S. if you find any little bugs when using it please fix them


machine_learning/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(machine_learning)
+
+find_package(catkin REQUIRED)
+
+# catkin_python_setup()
```
kev-the-dev: Do you need to uncomment this for python to work?

machine_learning/labeling/image_providers.py
```diff
@@ -0,0 +1,40 @@
...
+        - A video crawler
+        - Loading images from a folder
+    '''
+    name = "general image provider"
+
+    def load_source(self, *args, **kwargs):
+        ''' Load images from source '''
+        pass
+
+    def get_next_image(self):
```
kev-the-dev: might be nice to kwargs this too

kev-the-dev (changes requested):
Really helpful package. Please add setup.py and the catkin python setup thing so it's rosified

machine_learning/CMakeLists.txt
mattlangford: I wasn't planning on exposing any of the inner workings to the system, therefore it wouldn't need to be there.
The image provider class could be helpful externally, but I don't think the machine learning folder is where it should live. Until there's a nice home for it set up (maybe in `mil_dev_tools`), it should stay as a private file only used by the labeler.

machine_learning/labeling/image_providers.py
mattlangford: In reference to the above comment, only the labeler uses this function right now, so no need until it's exposed.

machine_learning/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+#   DEPENDENCIES
+#   std_msgs  # Or other packages containing msgs
+# )
+
+catkin_package(
+#  INCLUDE_DIRS include
+#  LIBRARIES machine_learning
+#  CATKIN_DEPENDS other_catkin_pkg
+#  DEPENDS system_lib
+)
```
DSsoto: Can you remove commented out boilerplate?

machine_learning/labeling/image_providers.py
```diff
@@ -0,0 +1,40 @@
...
+        ''' Returns the next image in the sequence '''
+        pass
+
+
+class RandomColors(BaseImageProvider):
+    name = "Random Color Generator"
```
DSsoto: I'm not seeing this being used anywhere, should it be a comment?

machine_learning/labeling/labeler.py
```diff
@@ -0,0 +1,248 @@
...
+
+# Tools ====================
+from tools import PaintBrushTool, PolygonTool, MoveTool
+
+tools = [PaintBrushTool, PolygonTool, MoveTool]
+
+
+'''
+A main interface for various labeling tools.
+'''
```
DSsoto: Is this a docstring for the whole labeler.py module?

machine_learning/labeling/labeler.py
```diff
@@ -0,0 +1,248 @@
...
+        if not os.path.exists(self.folder_path):
+            os.makedirs(self.folder_path)
+
+        self._name = "tool"
+        cv2.namedWindow(self._name, cv2.WINDOW_NORMAL)
+        cv2.setMouseCallback(self._name, self._mouse_cb)
+
+    def _generate_hotkeys(self, tools):
+        ''' Make hotkeys for each tool. Make sure each tool class has a unique name '''
+        if len(tools) == 0:
```
DSsoto: Like what you did there.

machine_learning/package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<package>
+  <name>machine_learning</name>
```
DSsoto: I think that just like with sub and boat, we should establish a  mil namespace. I propose that all packages in this repo start with `mil_`. What say you?

DSsoto (changes requested):
Requesting either package name change or good reasons not to, plus minor comments, questions.

Also have you considered making this a qt app?

machine_learning/package.xml
mattlangford: Sure

machine_learning/CMakeLists.txt
mattlangford: I like to leave in the lines that could possibly be used in the future, that way if someone adds a c++ file or wants messages, they can just uncomment the lines rather then going to some other cmake file and copying what they need

machine_learning/labeling/image_providers.py
mattlangford: Line 212 in the `labeler.py` file

machine_learning/labeling/labeler.py
mattlangford: I suppose it is

DSsoto (approved):

DSsoto:
This is ready to merge but shouln't be until **__The Great Merge__** is complete and this is rebased on top of it.

mattlangford:
Going to close based on some stuff. This'll still exist in my branch if we ever want it.

DSsoto:
Wait how come? This seems like pretty useful stuff. I just wanted to wait until the great merge was done since this is lower priority.

mattlangford:
@tessbianchi made some pretty compelling arguments to me about using an online labeling tool and adding infra to support that rather then rolling our own. The benefits of the online tool are too many to type so next time there is a call we can discuss.

DSsoto:
Oh yea, she talked to me about that too. Fair enough.

Issue #2, opened 2017-01-27
Create TO-DO list


kev-the-dev:
We should create a list of things to add / migrate to software-common

sentree:
We can either make a wiki page or use this issue to track the packages that need to be migrated. Any preference?

kev-the-dev:
Vision Lib: add any generic vision library functions from navigator + subjugator to the mil_vision_lib package

@DSsoto

kev-the-dev:
Drivers: Looks like there are some drivers / hardware interfaces that might be better off in the common repo. I see the roboteq motor controller drivers, xbox controller driver, and hydrophone / paulboard stuff as good candidates

kev-the-dev:
GNC: some gnc related things could be useful in generic versions, such as the OgridArbiter, MRAC (if we move all robot specific parameters to ros params), certain components of the selfcheckers in sub and navigator. @mattlangford

mattlangford:
I think the ogrid arbiter would be good to move (I can do that). MRAC is pretty specific to a boat (things were tuned to navigator), so I don't think that needs to be moved.
The self checks have to be tuned to each platform, I tried to make them general for navigator but there are a lot of things that need to be checked in special ways that aren't general enough - so I think it should stay.
Roboteq wasn't even our package, so no move. The Xbox controller driver wasn't ours, and the controller node was pretty specific to navigator, so I think it should stay.
Paulboard hydrophone stuff - for sure move, I bet @dsoto would be willing to do rhat

sentree:
This can be closed since #5 is being used to track the progress now.

PR #4, opened 2017-02-08
REPO: Perform initial setup of the software-common repository

sentree:
**This pull request is part of a series of four. All four are on different repositories and it is crucial that they are merged at the same time to keep the install process working!**

These are the other pull requests to look at:
* [installer #2](https://github.com/uf-mil/installer/pull/2)
* [Sub8 #199](https://github.com/uf-mil/Sub8/pull/199)
* [Navigator #192](https://github.com/uf-mil/Navigator/pull/192)

There are a few things that were done on most if not all of these repositories:
* Some outdated code and dependencies were removed
* Aliases were moved to a bash_aliases.sh file in the scripts directory
    * This is sourced by the ~/.milrc file now created by the install script
    * General aliases are here, specific ones stay with their respective projects
* All roads (or readme files, that is) now lead to the software common wiki (which I will be working on in the next few days)
    * Check out the vehicle images!

Now on to software-common specifics:
* The bag backup script was moved over and should run automatically once the sub OS is fully set up (definitely before Saturday)
* Common repository dependencies are now submodules instead of being pulled separately
    * Allows us to easily version control them (i.e. if we roll one back, that can be committed)
    * I recommend purging your workspace and running the install script after this goes through or there will be duplicate packages
* Two line bash is now a version controlled script, it will be installed next time the install script is run (I included a way to disable it, assuming I'm not the only one that doesn't like it [am I??])
* The ros_connect script is now an absolute beast
    * A hostname or IP can be passed with the -n argument to skip the ~2 seconds of host discovery
    * Persistence across shells is supported (run once and the master that is selected carries over to new shells)
    * Persistence can be toggled on the fly
    * One-time runs are supported with the -o argument (i.e. if I want just the current shell to have example.com as a ros master, not have it persist to new ones)
    * Yes, I left rsub and rnav as aliases for those stuck in their ways (ahem, @mattlangford)

### DIFF
--- a/scripts/bash_aliases.sh
+++ b/scripts/bash_aliases.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# This script will be sourced by the MIL runcom file if the install script
+# cloned this repository during a run or if the install script was run after
+# it was cloned manually. It contains general MIL commands that have been
+# aliased to make them faster or easier to execute. Becoming familiar with
+# these will likely increase productivity, so it is recommended to do so.
+
+# Source external alias modules
+source $CATKIN_DIR/src/software-common/scripts/two_line_bash.sh
+source $CATKIN_DIR/src/software-common/scripts/ros_connect.sh
+
+# Directory navigation
+alias swc="cd $CATKIN_DIR/src/software-common"
+
+# Bash sourcing
+alias srcbrc='source ~/.bashrc'
+
+# Catkin workspace management
+alias cm='catkin_make -C $CATKIN_DIR'
+
+# Alarms
+alias araise='rosrun ros_alarms raise'
+alias aclear='rosrun ros_alarms clear'
+alias areport='rosrun ros_alarms report'
+
+# Simulation
+alias killgazebo="killall -9 gazebo && killall -9 gzserver && killall -9 gzclient"
--- a/scripts/two_line_bash.sh
+++ b/scripts/two_line_bash.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Two line bash is a script that adds separators between commands to make
+# their output more readable. It also includes the current time and the
+# selected branch of the git repository the user is in. While it is enabled
+# by default, two line bash can be disabled by creating a .disable_tlb file in
+# the user's home directory. This can be accomplished with the following
+# command: touch ~/.disable_tlb
+
+function parse_git_branch {
+	PS_BRANCH=''
+	PS_FILL=${PS_LINE:0:$COLUMNS}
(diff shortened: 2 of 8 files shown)


sentree:
@DSsoto: IT'S HAPPENING **osprey*

sentree:
Once these are merged, I recommend that we rename a few of the repositories for the sake of standardization. These are just suggestions of course, but I thought I'd bring up the point:
* software-common -> mil_common
* rawgps-tools -> rawgps_tools
* Sub8 -> SubjuGator or subjugator
* Navigator -> NaviGator or navigator

scripts/two_line_bash.sh
```diff
@@ -0,0 +1,37 @@
...
+		return
+	elif [ -f _FOSSIL_ -o -f .fslckout ]; then
+		PS_BRANCH="(fossil $(fossil status|awk '/tags/{print $2}')) "
+		return
+	fi
+	ref=$(git symbolic-ref HEAD 2> /dev/null) || return
+	PS_BRANCH="(git ${ref#refs/heads/}) "
+}
+
+if [ ! -f ~/.disable_tlb ]; then
```
mattlangford, line 24: Ahem

scripts/bash_aliases.sh
```diff
@@ -0,0 +1,28 @@
...
+source $CATKIN_DIR/src/software-common/scripts/ros_connect.sh
+
+# Directory navigation
+alias swc="cd $CATKIN_DIR/src/software-common"
+
+# Bash sourcing
+alias srcbrc='source ~/.bashrc'
+
+# Catkin workspace management
+alias cm='catkin_make -C $CATKIN_DIR'
```
mattlangford, line 20: Can you add a -j8 in there too?

mattlangford (commented):
I like it! We're in the future.

I'll let a more bashy person approve this (@ironmig)

scripts/two_line_bash.sh
sentree: Relax, it's not the default!

scripts/bash_aliases.sh
sentree: I was under the impression that catkin_make automatically detected the number of available threads. Is this not the case?

DSsoto: Not the case.

DSsoto (approved):
Looks good, I have no qualms with any of this.

DSsoto:
I agree with @whispercoros  on the renaming of the repos, just to have standardized naming. I would also prefer SubjuGator to subjugator since it's a proper name. @mattlangford, thoughts?

PR #6, opened 2017-03-07
UTILS: Add online_bagger

kingkevlar:
This online_bagger utility is meant to allow us to retroactively bag data when something breaks during testing by continually streaming a predefined set of topics which would help troubleshoot. 

There is a file called 'online_bagger_example.yaml' that should be updated with the :
- topics 

- desired bag location (default to home/user/online_bagger)
- desired bag name (default to current date and time)
- desired bag time

The 'online_bagger_example.launch' file starts the 'online_bagger' node, 

Usage:
1. Update online_bagger_example.yaml' with topics to stream (add to list) and stream time (seconds)
2. Initiate node ('online_bagger') and service('/online_bagger/dump') :
    roslaunch mil_ros_tools online_bagger.launch


utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+            pass
+
+    def start_bagging(self, req):
+        '''
+        dumps all data in dictionary to bags, temporarily stops streaming
+        during the bagging process, resumes streaming when over
+        '''
+        self.dumping = req.data
+        self.streaming = False
+        bag = rosbag.Bag('ooka.bag', 'w')
```
RustyBamboo: Perhaps save the bag as the current time?

utils/mil_ros_tools/online_bagger/config/topics_to_stream.yaml
```diff
@@ -0,0 +1,13 @@
+# list an arbitrary set of topics to  subscribe to and stream:
+  # ram_limit is not yet implemented
```
RustyBamboo: https://docs.python.org/3/library/sys.html#sys.getsizeof

RustyBamboo (changes requested):

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
RustyBamboo: And perhaps add the path for bag folder in the yaml?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+import rospkg
+from std_srvs.srv import SetBool
+
+'''
+To call this service call the '/online_bagger' service with the SetBool Parameter set to True.
+Doing this will generate a rosbag with the last x number of seconds saved
+
+for example:
+
+    bagging_server = rospy.ServiceProxy('/online_bagger', SetBool)
```
mattlangford: It'd probably be good to make a service that allows you to pass a number in for the last number of seconds to dump

utils/mil_ros_tools/online_bagger/config/topics_to_stream.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+stream_time: 30 # seconds
```
mattlangford: I think if a RAM limit was put in place, time wouldn't matter

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+        self.streaming = True
+        self.dumping = False
+        # self.resubscribe
+
+        self.get_params()
+        self.make_dicts()
+        self.subscribe()
+
+
+        self.bagging_service = rospy.Service('/online_bagger', SetBool, self.start_bagging)
```
mattlangford: A more descriptive name would be good `/online_bagger/dump`

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+        self.ram_limit = rospy.get_param('/online_bagger/ram_limit')
+        self.stream_time = rospy.get_param('/online_bagger/stream_time')
+
+        rospy.loginfo('stream_time: %s', self.stream_time)
+        # rospy.loginfo('ram_limit: %s', self.ram_limit)
+
+    def make_dicts(self):
+        '''
+        make dictionaries with deques() that will be filled with topics
+        with all of the 5topics defined by the yaml
```
mattlangford: Typo

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+
+        rospy.loginfo('topics: %s', self.topic_list.keys())
+    def subscribe(self):
+        '''
+        Immediately initiated with instance of the BagStream class:
+        subscribes to the set of topics defined in the yaml configuration file
+        '''
+        # use failed topics list that can be resubscribed to at a later time?
+
+        for topic in self.topic_list.keys():
```
mattlangford: This will be important for robustness to act on these failed topics. The moment you create the subscriber objects, topics will asynchronously start flowing in, so you could loop here forever until all the topics are subbed to. The only other change would be moving the service provider line before `self.subscribe` was called.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+        if not(self.streaming):
+            return
+
+        self.n = self.n + 1
+        self.topic_list[topic].append((rospy.get_rostime(), msg))
+
+        time_diff = self.get_newest_topic_time(topic) - self.get_oldest_topic_time(topic)
+
+        # verify streaming is popping off and recording topics
+        # if self.n % 50 == 0:
```
mattlangford: `rospy.logdebug` would be good here. That would always log but never show up unless someone turned it on

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+        #     rospy.loginfo('time_diff: %s', time_diff.to_sec())
+        #     rospy.loginfo('topic: %s', topic)
+
+        if time_diff > rospy.Duration(self.stream_time):
+            self.topic_list[topic].popleft()
+        else:
+            pass
```
mattlangford: Is this necessary

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+
+        rospy.loginfo('dumping value: %s', self.dumping)
+        rospy.loginfo('bagging commencing!')
+
+        for topic in self.topic_list.keys():
+            rospy.loginfo('topic: %s', topic)
+            for msgs in self.topic_list[topic]:
+
+                bag.write(topic, msgs[1], t=msgs[0])
+                # rospy.loginfo('topic: %s, message type: %s, time: %s', topic, type(msgs[1]), type(msgs[0]))
```
mattlangford: This would be good to have

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,162 @@
...
+        50th message accross all topics.
+        
+        stream, callback function does nothing if streaming is not active
+        '''
+
+        if not(self.streaming):
+            return
+
+        self.n = self.n + 1
+        self.topic_list[topic].append((rospy.get_rostime(), msg))
```
mattlangford: Might be a good idea to check if the message has a header and use that as the time if it exists. 
```python
if hasattr(msg, 'header'):
    time = msg.header.stamp
else:
    time = rospy.get_rostime()
```

mattlangford (changes requested):
Really good stuff, there are some important things to change before merging.
Also a useful addition would be a status topic: how many topics are currently subscribed to, what's the message count, the memory usage, etc.

utils/mil_ros_tools/online_bagger/scripts/online_bagging_client.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
```
mattlangford: This file is probably not needed as well

utils/mil_ros_tools/online_bagger/config/topics_to_stream.yaml
```diff
@@ -0,0 +1,13 @@
...
+      # down_stream_rate: 10 Hz 
+ram_limit: 0.5 # gb 
+message_topics: ["/odom",
```
kev-the-dev: just "topics " would probably be a better name for this parameter.

Also, because this is going in the common repo, perhaps this config and the launch file should be renamed to "online_bag_example.yaml and online_bag_example.launch" as individual projects will probably create their own configs.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
kev-the-dev: I agree with Matt. I think two services would be nice, one std_srvs/Trigger service to dump all buffered data and a custom service for dumping x seconds (perhaps also with the ability to specify filename for labeling purposes)

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+        self.make_dicts()
+        self.subscribe()
+        rospy.loginfo('subscriber success: %s', self.subscriber_list)
+
+
+        self.bagging_service = rospy.Service('/online_bagger/dump', SetBool, self.start_bagging)
+
+    def get_params(self):
+
+        self.topics = rospy.get_param('/online_bagger/message_topics')
```
kev-the-dev: generally parameters are not in the global name space (starting with /) like this. If you just put the parameter name (like "topics" or "stream_time"), ros will known that you are talking about the param in this node's name space)

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+        Each element in self.subscriber list is a list [topic, Bool]
+        where the Bool tells the current status of the subscriber (sucess/failure)
+        '''
+
+        for topic in self.subscriber_list:
+            if topic[1] is True:
+                pass
+            else:
+                msg_class = rostopic.get_topic_class(topic[0])
+                if msg_class[1] is None:
```
kev-the-dev: This would be a good place to log that there was an error and perhaps set up a rospy.Timer to retry at an interval.

kev-the-dev (commented):
I'm really exited to have this tool. Request some refactoring with the bagging service(s) and some ros standards stuff.

utils/mil_ros_tools/online_bagger/config/topics_to_stream.yaml
```diff
@@ -0,0 +1,13 @@
+# list an arbitrary set of topics to  subscribe to and stream:
+  # ram_limit is not yet implemented
+
+stream_time: 30 # seconds
+      # down_stream_rate: 10 Hz 
+ram_limit: 0.5 # gb
```
sentree: Once this is implemented, I recommend some sort of check to make sure that this is not over the maximum available RAM. I would perhaps even go a step further and leave a percentage free, but that's up to you.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class OnlineBagger(object):
+    def __init__(self):
+        '''
+        - makes dictionary of dequeues
+        - subscribes to set of topics defined by the yaml file in directory
+        - have a streaming option that can be turned on and off in the callback
+        - option to bag data
+        - add a re subscribing option for anything that failed? maybe...?
```
sentree: This appears to be implemented, so why the open question?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Each element in self.subscriber list is a list [topic, Bool]
+        where the Bool tells the current status of the subscriber (sucess/failure)
+        '''
+
+        for topic in self.subscriber_list:
+            if topic[1] is True:
+                pass
+            else:
+                msg_class = rostopic.get_topic_class(topic[0])
```
sentree: I try not to use pass too much, there are usually alternatives:

            if topic[1] is not True:
                msg_class = rostopic.get_topic_class(topic[0])
                ...

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+                if msg_class[1] is None:
+                    pass
+                else:
+                    rospy.Subscriber(topic[0], msg_class[0], lambda msg, _topic=topic[0]: self.callback(msg, _topic))
```
sentree: Same thing here:

                if msg_class[1] is not None:
                    rospy.Subscriber(topic[0], msg_class[0], lambda msg, _topic=topic[0]: self.callback(msg, _topic))
                    ...

utils/mil_ros_tools/online_bagger/scripts/online_bagging_client.py
sentree: I agree, this should be in whatever program is using online_bagger

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+        '''
+        return self.topic_list[topic][0][0]
+
+    def get_newest_topic_time(self, topic):
+        '''
+        returns newest timestamp for a given topic
+        '''
+        return self.topic_list[topic][-1][0]
+
+    def callback(self, msg, topic):
```
sentree: I would name this something more specific. When you're dealing with many callbacks, it will be easier to tell them apart.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,188 @@
...
+        if self.n % 50 == 0:
+            rospy.loginfo('time_diff: %s', time_diff.to_sec())
+            rospy.loginfo('topic: %s', topic)
+
+        if time_diff > rospy.Duration(self.stream_time):
+            self.topic_list[topic].popleft()
+        else:
+            pass
+
+        self.subscribe()
```
sentree: I recommend logging the topics that could not be subscribed to.

sentree (commented):
Only one commit. Sad! History aside, this is solid work. I have some questions, a couple of suggestions, on how to merge instead of being closed.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
mattlangford: Filename would be a good idea. It could still just be one service but calling it with blank time field would make it dump everything

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,400 @@
...
+        if False: topic still needs to be subscribed to and
+        subscriber will be run.
+
+        Each element in self.subscriber list is a list [topic, Bool]
+        where the Bool tells the current status of the subscriber (sucess/failure).
+
+        Return number of topics that failed subscription
+        """
+
+        i = 0  # failed subscriptions
```
mattlangford: In the future, it'd be best to just name this variable `failed_subscription_count` or something like that - nobody likes too many `i`'s.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,400 @@
...
+
+        return len(self.topic_messages[topic])
+
+    def get_total_message_count(self):
+
+        """
+        Returns total number of messages across all topics
+        """
+
+        self.total_message_count = 0
```
mattlangford: Does this need to be a `self` variable?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,400 @@
...
+        self.set_bag_directory()
+
+        rospy.loginfo('dumping value: %s', self.dumping)
+        rospy.loginfo('bagging commencing!')
+
+
+        for topic in self.topic_messages.keys():
+            rospy.loginfo('topic: %s', topic)
+            if not req.amount:
+                self.i = 0
```
mattlangford: Don't want more variables in the `self` space than needed

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,400 @@
...
+
+        return self.ram_usage
+
+    def set_ram_limit(self, topic):
+
+        """
+        Limit ram usage by removing oldest messages
+        """
+
+        if self.get_ram_usage() > self.ram_limit:
```
mattlangford: This won't really enforce the limit will it? If there's a topic that initially got small sized messages but now get's bigger messages the deleting will have no effect and lead to messages that don't need to be deleted being lost, or a topic that isn't published very frequently that contains big messages that hog all the RAM and rarely get's cleared.
New proposition: per topic RAM limits and using `sys.getsizeof(...)` to detect violations to the deque size. This will prevent losing small important data and make sure everything is as used as possible.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self):
+
+        """
+        Make dictionary of dequeues.
+        Subscribe to set of topics defined by the yaml file in directory
+        Stream topics up to a given ram limit, dump oldest messages when limit is reached
+        Set up service to bag n seconds of data default to all of available data
+        """
+
+        self.i = 0  # successful subscriptions
```
mattlangford: Might call this `successful_subscription_count` and `self.n` could be `iteration_count` or something more descriptive

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+
+        self.topics = rospy.get_param('/online_bagger/topics')
+        self.ram_limit = rospy.get_param('/online_bagger/ram_limit') * 1.073e9  # gigabytes to bytes
+        # self.stream_time = rospy.get_param('/online_bagger/stream_time')
+        self.dir = rospy.get_param('/online_bagger/bag_package_path')
+        self.bag_name = rospy.get_param('/online_bagger/bag_name')
+
+        if not self.bag_name:
+            self.bag_name = str(datetime.date.today()) + '-' + str(datetime.datetime.now().time())[0:8]
```
mattlangford: Will this permanently set the name of the bag to the date when the program is started?

utils/mil_ros_tools/online_bagger/srv/BaggerCommands.srv
```diff
@@ -0,0 +1,4 @@
+float32 amount
+string unit
```
mattlangford: Weren't you going to add a `name` field

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+    def subscribe_loop(self):
+
+        """
+        Continue to subscribe until at least one topic is successful
+        """
+
+        i = 0
+        while not self.subscribe():
+            i = i + 1
+            self.subscribe
```
mattlangford: Typo?
Also, for any python ROS while loop, it's always good to have this stuff: 
```python
while conditions and not rospy.is_shutdown():
    # logic
    # logic
    rospy.sleep(0.1)  # So it doesn't hog all the cpu
```

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+        for topic in self.subscriber_list:
+            if topic[1] == False:
+                continue
+
+            topic = topic[0]
+            rospy.loginfo('topic: %s', topic)
+
+            # if no number is provided or a zero is given, bag all messages
+            if not req.amount:
+                self.i = 0
```
mattlangford: Is this the same `self.i` used by `self.subscribe`?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+    def start_bagging(self, req):
+
+        """
+        Dump all data in dictionary to bags, temporarily stops streaming
+        during the bagging process, resumes streaming when over.
+        """
+
+        rospy.loginfo('bagging %s %s', req.amount, req.unit)
+
+        self.dumping = True
```
mattlangford: This never get's changed from `True` does it? Does it get used anywhere else?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+        rospy.loginfo('failed topics: %s', self.subscriber_list)
+
+    def subscribe_loop(self):
+
+        """
+        Continue to subscribe until at least one topic is successful
+        """
+
+        i = 0
+        while not self.subscribe():
```
mattlangford: Will this ever run? `self.i` will always be > 0 as soon as one topic is subscribed to.

mattlangford (changes requested):
Some flow issues
I think the final big change is to change the way the RAM limit is imposed - I think that's something that's pretty important to get right

utils/mil_ros_tools/online_bagger/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(mil_ros_tools)
+
+find_package(catkin REQUIRED COMPONENTS
+  rospy
+  message_generation
+  
+)
+add_service_files(
```
DSsoto: Looks good, but this will need to change when you rebase onto the great merge branch and make this part of the mil_tools pkg, don't forget to add message_generation to it.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Return number of topics that failed subscription
+        """
+
+
+        for topic in self.subscriber_list:
+            if not topic[1]:
+                msg_class = rostopic.get_topic_class(topic[0])
+                if msg_class[1] is None:
+                    pass
```
DSsoto: Can you make this `if msg_class[1] is not None:` and then the block after the `else`?

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+
+        self.usage = resource.getrusage(resource.RUSAGE_SELF)
+
+        # getattr returns ram usage in kilobytes multiply by 1e3 to obtain megabytes
+        self.ram_usage = getattr(self.usage, 'ru_maxrss')*1e3
+
+        return self.ram_usage
+
+    def set_ram_limit(self, topic):
```
DSsoto: You're not setting the ram limit here, you're enforcing it.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+
+        """
+        Limit ram usage by removing oldest messages
+        """
+
+        if self.get_ram_usage() > self.ram_limit:
```
DSsoto: I would put replace the word `if` with `while`, so the limit actually enforced.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+
+        topic_duration = self.get_topic_duration(topic).to_sec()
+
+        ratio = requested_seconds / topic_duration
+        index = math.floor(self.get_topic_message_count(topic) * (1 - min(ratio, 1)))
+
+        self.bag_report = 'The requested %s seconds were bagged' % requested_seconds
+
+        if not index:
```
DSsoto: As a matter of style, I think it's always nicer to check for equality with zero for numerical quantities.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,427 @@
...
+            self.bag_report = 'Only %s seconds of the request %s seconds were available, all \
+            messages were bagged' %  (topic_duration, requested_seconds)
+        return int(index)
+
+    def check_action(self, req):
```
DSsoto: I think you should just accept a single unit, whether minutes or seconds. That way you can remove the unit field for the msg and it will be quicker to call, and less error prone.

DSsoto (commented):
Good work! After this is rebased hopefully we can merge and you can ascend to bigger and better things.

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,332 @@
...
+            messages were bagged' %  (topic_duration, requested_seconds)
+        return index
+
+    def set_bagger_status(self):
+
+        """
+        Print status of online bagger
+        """
+
+        self.bagger_status = ('Subscriber List: ' + str(self.subscriber_list)  + ' Message Count: '
```
mattlangford: You can put a `\` at the end of the first line and not need `( )`'s

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
```diff
@@ -0,0 +1,332 @@
...
+        time_diff = self.get_topic_duration(topic)
+
+        # verify streaming is popping off and recording topics
+        if self.iteration_count % 100 == 0:
+            rospy.logdebug('time_difference: %s', time_diff.to_sec())
+            rospy.logdebug('topic: %s', topic)
+            rospy.logdebug('topic type: %s', type(msg))
+            rospy.logdebug('number of topic messages: %s', self.get_topic_message_count(topic))
+
+        if time_diff > rospy.Duration(self.stream_time):
```
mattlangford: Should be `while`, right? (and update `time_diff` in the body)

mattlangford: Also, I think it's important to be able to specify stream time's on a per topic basis - i.e. we might be okay with 5 minutes of odom but only 20 seconds of images.

mattlangford (dismissed):
The while loop is important and I think each topic needs to have it's own stream time

utils/mil_ros_tools/online_bagger/nodes/online_bagger.py
kingkevlar: So I was thinking about putting it in a while loop until I realized that all the callbacks are processed in series so as long as the frequency of the messages don't change (which might not be a good assumption) it's impossible for the messages to ever get more than one time step ahead of the stream time. So popping g a single message off will not be able to cause a drift in stream time the way it could drift with ram usage (because message can differ in size). I guess what I'm wondering is: if /odom randomly drops out for 1.0 seconds and we get it back (is that even a possible scenario?), do we actually want to drop the last 100 messages when we know we haven't actually increased the size of our dequeue?

For individual stream topic times, should I change the client . srv file to a percent based bagging amount (ie. Bag 50% of available data)? That way if we want to bag everything it's a lot cleaner than checking how many seconds are available on each topic. Or something else if you have a better idea.

Or have it take in a list since right now it can bag the last n seconds of available data across all topics.

Or maybe I should follow through with the idea @ironmig  had which was have two separate services, one that bags everything and one that is more custom, so that we can call the service on the fly quickly if needed, or is that too much trouble to keep track of two services?

kingkevlar: Oh wait, my comment about losing /odom is no good if we lose /odom for 1.0 seconds before hitting the limit of the stream time.

kingkevlar: @DSsoto said that if we specify stream times on a per topic basis then when we call the service it should be to bag everything. I agree with that sentiment, so that when we call the service it only takes in the bag name and doesn't get any messier.

I'll have the service default to a stream time set in the beginning of the yaml, you will be able to override on a per topic basis if you desire.  

Also, I'd really like to start working on generating the point clouds from the imaging sonar... which seems more important at the moment...

mattlangford: What happened to the idea of having one large ring buffer - that would solve all the problems I think. Dumping from a specific time would be easy since you'd only have to search through one list and keeping the memory limit would be easy since you'd just non-discriminately purge from the end. It would naturally fill up on the small messages and (almost) always stay under the memory limit.

kingkevlar: I'm going to make this in C++, I'll get back to you when I'm done. (Give me a month or two)

PR #7, opened 2017-03-09
SCRIPTS: Add wsmux and bmux

sentree:
References to software-common were updated to point to mil_common and the rawgps-tools submodule was replaced with odometry_tools.

Some general script cleanup was done as I noticed minor style discrepancies. Nothing major was changed in `scripts/bash_aliases.sh`, `scripts/bag_backup.sh`, or `scripts/ros_connect.sh`.

Two new functions were added to bash to make operating vehicles on testing days easier. Both of these are well documented in their respective files and help commands, so this will only provide a cursory overview.

The wsmux function allows users to switch between catkin workspaces quickly (autocompletion is magic) so that each user can have their own workspace as opposed to their own user account. This will eliminate all of the issues we had with user account permissions.

The bmux function was written in response to @DSsoto's request in [NaviGator #192](https://github.com/uf-mil/NaviGator/pull/192). It allows the user to source, list, and clear the bagging aliases for a specific vehicle repository. This will prevent loading a bunch of variables into user's local shells; however, it will allow them to load the aliases if they choose to and will allow them to be loaded automatically on the actual vehicles. These variables (see the `scripts/bagging_variables.sh` file in [NaviGator #193](https://github.com/uf-mil/NaviGator/pull/193)) will be used with the bag function, which will be included in a future pull request.

### DIFF
--- a/scripts/ros_connect.sh
+++ b/scripts/ros_connect.sh
@@ -196,12 +196,6 @@ ros_disconnect() {
 }
 
 
-# Prints debugging output for the master roscore that is currently selected
-alias ros_env='echo "ROS_IP=$ROS_IP
-ROS_HOSTNAME=$ROS_HOSTNAME
-ROS_MASTER_URI=$ROS_MASTER_URI"'
-
-
 # Generates the persistence file if it does not exist
 if [ ! -f $PERSISTENCE_FILE ]; then
 	echo $DEFAULT_HOST > $PERSISTENCE_FILE
--- a/scripts/ros_connect.sh
+++ b/scripts/ros_connect.sh
@@ -102,29 +102,15 @@ ros_connect() {
 	while [ "$#" -gt 0 ]; do
 		case $1 in
 			-h|--help)
-				echo "Usage: ros_connect [OPTION] [HOSTNAME]..."
+				echo "Usage: ros_connect [OPTION]..."
 				echo "Manager for connections to remote roscores."
 				echo ""
 				echo "Option		GNU long option		Meaning"
 				echo "-h		--help			Display the help menu"
-				echo "-p		--persistence		Toggle persistence across shells"
-				echo "-o		--one-time		Only set the roscore for this shell"
 				echo "-n [HOSTNAME]	--hostname		Manually pass in a hostname"
+				echo "-o		--one-time		Only set the roscore for this shell"
+				echo "-p		--persistence		Toggle persistence across shells"
 				HOST_DISCOVERY=false
-				PERSIST=false
-				shift 1
-				;;
-			-p|--persistence)
-				if [ -z `cat $PERSISTENCE_FILE | grep "disabled"` ]; then
-					echo "disabled" > $PERSISTENCE_FILE
-				else
-					echo $DEFAULT_HOST > $PERSISTENCE_FILE
-				fi
-				HOST_DISCOVERY=false
-				PERSIST=false
-				shift 1
-				;;
-			-o|--one-time)
 				PERSIST=false
 				shift 1
 				;;
@@ -138,6 +124,20 @@ ros_connect() {
 				set_ros_master $HOST
 				HOST_DISCOVERY=false
 				shift 2
+				;;
+			-o|--one-time)
+				PERSIST=false
+				shift 1
+				;;
+			-p|--persistence)
+				if [ -z `cat $PERSISTENCE_FILE | grep "disabled"` ]; then
+					echo "disabled" > $PERSISTENCE_FILE
+				else
+					echo $DEFAULT_HOST > $PERSISTENCE_FILE
+				fi
+				HOST_DISCOVERY=false
+				PERSIST=false
+				shift 1
 				;;
 			*)
 				echo "Option $1 is not implemented."
--- a/scripts/ros_connect.sh
+++ b/scripts/ros_connect.sh
@@ -11,6 +11,7 @@
 
 
 # These parameters define the network to search on
+DEFAULT_HOST="localhost"
 SEARCH_DOMAIN="ad.mil.ufl.edu"
 SUBNET="192.168.37.0/24"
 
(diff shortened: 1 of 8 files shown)


scripts/ros_connect.sh
```diff
@@ -105,42 +102,42 @@ ros_connect() {
...
-				echo "Usage: ros_connect [OPTION] [HOSTNAME]..."
+				echo "Usage: ros_connect [OPTION]..."
 				echo "Manager for connections to remote roscores."
 				echo ""
 				echo "Option		GNU long option		Meaning"
 				echo "-h		--help			Display the help menu"
-				echo "-p		--persistence		Toggle persistence across shells"
-				echo "-o		--one-time		Only set the roscore for this shell"
 				echo "-n [HOSTNAME]	--hostname		Manually pass in a hostname"
+				echo "-o		--one-time		Only set the roscore for this shell"
```
DSsoto, line 111: I like this, good stuff.

scripts/ros_connect.sh
```diff
@@ -105,42 +102,42 @@ ros_connect() {
...
 			-o|--one-time)
 				PERSIST=false
 				shift 1
 				;;
-			-n|--hostname)
-				HOST="$2"
-				if [ "$HOST" != "localhost" ]; then
-					set_ros_ip
+			-p|--persistence)
+				if [ -z `cat $PERSISTENCE_FILE | grep "disabled"` ]; then
```
DSsoto, line 133: _**Suggestion** (but not for this PR):_  Instead of cluttering the user's home directory with empty configuration files, you could make a variable at the top of the script that stores the persistence setting (boolean). To change it, you would find and replace the current value in the script itself and then re-source this script.

scripts/ros_connect.sh
```diff
@@ -198,11 +195,18 @@ ros_disconnect() {
 	ros_connect -n $DEFAULT_HOST
 }
 
+
 # Prints debugging output for the master roscore that is currently selected
 alias ros_env='echo "ROS_IP=$ROS_IP
```
DSsoto: I feel like this should be in the bash aliases file.

scripts/two_line_bash.sh
```diff
@@ -21,17 +29,25 @@ function parse_git_branch {
...
-	PS_INFO="$GREEN\u@\h$RESET:$BLUE\w"
+	PROMPT_COMMAND=parse_tlb_info
+	PS_LINE=`printf -- "- %.0s" {1..200}`
+	PS_FILL=${PS_LINE:0:$COLUMNS}
+	PS_INFO="$GREEN\u@\h$RESET:$BLUE\w "
+	PS_CATKIN="$YELLOW\$PS_WORKSPACE"
 	PS_GIT="$YELLOW\$PS_BRANCH"
 	PS_TIME="\[\033[\$((COLUMNS-10))G\] $RED[\t]"
-	export PS1="\${PS_FILL}\[\033[0G\]${PS_INFO} ${PS_GIT}${PS_TIME}\n${RESET}\$ "
+	export PS1="\${PS_FILL}\[\033[0G\]${PS_INFO}${PS_CATKIN}${PS_GIT}${PS_TIME}\n${RESET}\$ "
```
DSsoto, line 52: `${PS_CATKIN}` should not be part of the prompt by default. There should be a way to include it on machines that actually have multiple workspaces.

scripts/wsmux.sh
```diff
@@ -0,0 +1,103 @@
...
+	# Check for workspaces in the home directory based on the argument
+	for FILE in ~/"$2"*; do
+
+		# Skip any directory that does not contain the marker file
+		[[ -f $FILE/$WS_MARKER ]] || continue
+
+		# Append just the name of the workspace folder to the autocomplete list
+		COMPREPLY+=( `echo "$FILE" | rev | cut -d "/" -f1 | rev` )
+	done
+}
```
DSsoto, line 26: I'm not a fan of leaving marker files scattered around. If there is another alternative, I would be happy to see it explored. For now, I'm ok giving this a shot.

DSsoto (changes requested):
Requesting that ${PS_CATKIN} is not added to the prompt by default. Everything else is a suggestion.

scripts/ros_connect.sh
sentree: I feel like that's not exactly the best way to do this. Rewriting and resourcing the program every time it is run just seems like a bad idea. I agree that keeping them in the home directory is probably not the best idea, but as far as I can tell, that is what most programs do. Perhaps I should move them to a `~/.uf-mil` folder at some point?

scripts/ros_connect.sh
sentree: Noted

scripts/two_line_bash.sh
sentree: Are you suggesting some logic to detect the number of workspaces or a setting in a configuration file somewhere?

scripts/wsmux.sh
sentree: That marker file is actually generated by Catkin. I used it for detection purposes because it was simple and already implemented.

scripts/ros_connect.sh
DSsoto: You didn't provide a reason for why it would be a bad idea. But since we already have a ~/.milrc, it seems like a ~/.uf-mil/ could be a clean way to store all this stuff.

scripts/wsmux.sh
DSsoto: Okay

scripts/ros_connect.sh
sentree: It would likely work; however, there could be glitches that weren't considered that result in corruption of the program. It is also far too much overhead for a simple bash function. If everyone agrees, I can move things to `~/.mil` (matches up better with `~/.milrc`) in the next pull request.

scripts/two_line_bash.sh
DSsoto: Maybe there should be a config file in ~/.uf-mil/ that stores simple settings such as this.

scripts/ros_connect.sh
DSsoto: I'd like that better, we can leave that for the next PR.

DSsoto (approved):

Issue #5, opened 2017-02-24
File migration list (sweet, sweet, great, great merge)


kev-the-dev:
Please list all files you want migrated from sub8 or navigator to the software-common repo. List the full path of where the file is from, and where it should go in the new repo. If the file is going to be a merge of sub8 and navigator files, say so. 

For now we are just worrying about getting the files in the repo. We will go through and refactor later, but feel free to leave notes on what needs to be done.

kev-the-dev:
# TOOLS

**sub8**
utils/sub8_ros_tools/sub8_misc_tools/text_effects.py
utils/sub8_ros_tools/sub8_ros_tools/geometry_helpers.py
utils/sub8_ros_tools/sub8_ros_tools/image_helpers
utils/sub8_ros_tools/sub8_ros_tools/init_helpers.py
utils/sub8_ros_tools/sub8_ros_tools/msg_helpers.py
utils/sub8_ros_tools/sub8_ros_tools/threading_helpers.py
utils/sub8_ros_tools/sub8_ros_tools/__init__.py
utils/sub8_ros_tools/setup.py
utils/sub8_ros_tools/package.xml 


**navigator**
utils/navigator_tools/nodes/clicked_point_recorder.py
utils/navigator_tools/nodes/tf_fudger.py
utils/navigator_tools/nodes/video_player
utils/navigator_tools/nodes/ogrid_draw.py
utils/navigator_tools/nodes/tf_to_gazebo.py
utils/navigator_tools/nodes/navigator_tools/cv_debug.py
utils/navigator_tools/nodes/navigator_tools/geometry_helpers.py

***new sub8***
utils/sub8_ros_tools/CMakeLists.txt
utils/sub8_ros_tools/sub8_ros_tools/func_helpers.py
utils/sub8_ros_tools/sub8_misc_tools/download.py
utils/sub8_ros_tools/test_ros_tools/test_ros_tools.py
utils/sub8_misc_tools/__init__.py
utils/sub8_ros_tools/sub8_tools/__init__.py

***new navigator***
utils/navigator_tools/navigator_tools/cv_debug.py
utils/navigator_tools/navigator_tools/rviz_helpers.py
utils/navigator_tools/navigator_tools/geometry_helpers.py
utils/navigator_tools/navigator_tools/msg_helpers.py
utils/navigator_tools/set.py
utils/navigator_tools/navigator_tools/__init__.py
utils/navigator_tools/package.xml
[...]

PR #8, opened 2017-03-09
REPO: The Great Merge

sentree:
This is it. The moment we've all been waiting for. Sit back and enjoy the crisp, clean history that is mil_common. Bask in the glow of those sweet, sweet merge commits.

Okay, enough basking. Let's get down to business. I think it's important to document how this small miracle was pulled off. After a lot of research, I was led to [this](https://stackoverflow.com/questions/19954485/extract-multiple-directories-using-git-filter-branch) Stack Overflow question. Basically, the command in the accepted answer removes everything except the chosen files and directories from the repository, then prunes the commit history to remove any unrelated files and unnecessary commits. What we're left with is a clean history based only on the chosen files and directories.

This command is the one that was used for that:
`git filter-branch -f --prune-empty --index-filter 'git rm --cached -r -q -- . ; git reset -q $GIT_COMMIT -- <FILES_TO_KEEP>' -- --all`

Once SubjuGator and NaviGator only contained what we wanted to move (specified in #5), it was just a matter of setting up remotes for each of them in mil_common and pulling from them one at a time, creating a single merge commit for each repository. There were no merge conflicts, but I suspect that is a product of folders being prefixed with `sub8` or `navigator`.

Now we move on to the fun part - stripping all of the files and directories of their namespaces and making them common. My intention was to let each user that specified files in #5 do this for the ones they added to the list. They probably understand that code better than I do and this will prevent us from stepping on each other's feet. Feel free to use that issue to delegate files if you want someone else to move them (namely me).

Once we get all of that done, we will need to merge this branch back into master and change references to and delete the old packages in SubjuGator and NaviGator respectively. Then we will officially have no overlaps in vehicle repository code!


kev-the-dev:
As long as there are no files missing, let's merge this so we can start fixing it

mattlangford:
Realized that I had only half of the review done and never submitted it. Those fixes will be fun...

DSsoto:
@mattlangford The real review will happen when we merge this to master.

sentree:
Let's try to split the reviews up for each pull request implementing fixes. As much as I'd love to have the largest review in history here, I think it will be simpler to look at things package by package.

PR #9, opened 2017-03-14
TOOLS: Fix mil tools from great merge

mattlangford:
First set of fixes from great merge.
It looks like a lot of files are missing, including most of the CMakes...

I'll probably go through and update some of the actual files that used `sub8_tools` to point to `mil_tools`


scripts/two_line_bash.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+#
+#
+# This was casuing issues in tmux and always showed an "catkin_mil_ws", I'm sure that was intended
```
mattlangford, line 5: Tmux didn't behave correctly after going full screen and back - @whispercoros can you explain why the catkin mil_ws was added?

utils/mil_tools/nodes/tf_fudger.py
```diff
@@ -122,11 +122,12 @@ def reset():
     if k == ord('z'):
         set_bars((0, 0, 0), (0, 0, 0))
 
-    if k == ord('s'):
+    if False and k == ord('s'):
```
mattlangford, line 125: This part of the script may be removed, else you'd have to specify the path for the launch file

scripts/two_line_bash.sh
sentree: This shows the currently sourced catkin workspace. It is intended to compliment wsmux since we are using different workspaces as opposed to user accounts on vehicles going forward.

I did not realize so few people had only a single workspace until @DSsoto brought it up, so I will be adding logic to only show this when there are multiple workspaces in `~/`. As for the tmux issues, I will look into that once I have made my changes.

I do not recommend altering this file right now or there will be many a merge conflict with my fork. If you insist, though, do it on `master` as these changes were not a part of ***The Great Merge***.

sentree:
If any files are missing, it is because they were not listed in #5. If you want to include them, it's not too late for me to roll back the merge and execute it again with those files in it. Just let me know on #5.

mattlangford:
I don't know which files are specifically missing, but I came across 5 when doing this PR

scripts/two_line_bash.sh
mattlangford: Sounds good, will fix in the morning

sentree: I found an issue with the number of dashes being printed after changing the terminal's width, but nothing weird with respect to tmux. Can you explain the issue in a bit more detail?

mattlangford: When I went to full screen (control-b z) the prompt would expand to fill the whole screen. When I'd do the same command to exit full screen, it would keep the length of dashes the same which would make it like a 5 line prompt until I cleared the screen.

sentree: That's the issue I caught last night, can't believe I missed it. If you want to check, you could pull the file in #10, but I'm sure it works now.

kev-the-dev:
I updated #5 with the missing tools files Matt or I found. I think it would be easier to just bring the entire utils/navigator_tools and utils/sub8_ros_tools folders from their respective repos into the project, then we can remove any files we aren't going to keep

mattlangford:
Covered in another PR

PR #10, opened 2017-03-14
SCRIPTS: Add bag and clean up configuration files

sentree:
**NOTE:** Please merge [installer #5](https://github.com/uf-mil/installer/pull/5) before this.

Together, the bmux and bag commands will help streamline bagging for test days and make it less of a chore. Bagging variables are configured on a per-vehicle basis in a `scripts/bagging_variables.sh` file. See the one currently in the NaviGator repository for an example. The special BAG_DIR and BAG_ALWAYS variables specify where to store bags and topics that should be included in every bag. All other variables should be prefixed with `bag_`.

Once the aliases are loaded by bmux, they can be used with bag. The command supports autocompletion of aliases, passing extra arguments to rosbag, and even on the fly grouping (for example, creating a temporary torpedo_board variable from other variables and topics). It also *strongly encourages* users to name bags, so we shouldn't have to go on bag treasure hunts anymore.

Two line bash will now only print the sourced catkin workspace if there are more than one in `~/`.

Configuration files are now stored in one central location, under `~/.mil`. For those of you following along at home, this means you should delete `~/.disable_tlb` and `~/.ros_connect_persistence` after running the install script (if they exist on your system).

### DIFF
--- a/scripts/bag.sh
+++ b/scripts/bag.sh
@@ -1,0 +1,153 @@
+#!/bin/bash
+
+# The bag command works in tandum with the bmux command. Once the bagging
+# variables have been loaded with bmux, the bag command can be used with
+# autocompletion to select topics. It is mainly a convenience layer for
+# rosbag; however, it does introduce a few features. A list of BAG_ALWAYS
+# topics can include topics that every bag should contain, bag directory
+# management is handled cleanly, and naming bags is... strongly encouraged.
+
+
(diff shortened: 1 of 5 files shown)


sentree:
I would also like to address this so called "two line bash issue" with the failing tmux (see #9) in this pull request, so hold off on merging.

DSsoto (approved):
Cool scripts breh

PR #12, opened 2017-03-17
TOOLS: import tools from navigator and sub8

kev-the-dev:
- move all platform-agnostic tools from sub8 and navigator to a new package, mil_dev_tools
- merge any conflicts from common tools in sub8 and navigator
- change namespaces / imports in mil_dev_tools to reflect this change
- temporarily move all remaining great merge files to .great_merge to keep things neat and have catkin not try to build it (maybe?)
- Add gitignore from #9 
- Langford the Langford who Lanforded me
- _ow_


mattlangford:
Could you explain the motivation for this? There's not a need to copy over Navigator stuff into this and then just delete all the lines from all the files. The tests could be copied back but I don't think they are necessary. I was trying to move away from the `import module` style of imports from the init files and prefer the `from module import *` since it lessens the clutter in actual files. `mil_tools` is shorter than `mil_dev_tools` and that's something that will be written a lot.
I think changing the name of all the packages is a good idea.

kev-the-dev:
The github diff doesn't show it well, but this is the merging of the two histories in sub8 and navigator so all history is preserved which is the intention of the great merge. This ensured that any differences were manually resolved to keep all functions in both. @DSsoto and others in MIL preferred dev tools because tools is too broad (e.g there may be vision tools also). I don't really have an opinion on python import convention so happy to change it.

mattlangford:
I'd prefer having all tools in one place accessible via `mil_tools`. In my PR I have `mil_ros_tools`, `mil_misc_tools`, (and `mil_perception_tools` could be added) all separate. Each of those can be imported on their own but if you want access to all the tools with one import, `mil_tools` is how you get it. 
It seems like having perception tools all by themselves in a sub directory of the perception folder doesn't seem to make sense to me and should actually be moved here. Plus in the `nodes` folder, there lives run time-y tools, not really `dev`elopment tools.

Also, there's no reason to move all the Navigator stuff here just to remove it - all that history is preserved in the Navigator repo where is belongs.

kev-the-dev:
Ya I agree bout the package name. There were some cases where new functions added in navigator's version of tools were added to the common package, so good to have that history. Again, the diff doesn't show it but this is the merging of the two tools histories which have some differences that were resolved.

PR #13, opened 2017-03-21
Import perception, uf-common, drivers

kev-the-dev:
@DSsoto's work to begin importing vision, drivers, and uf-common messages.


mattlangford (commented):
Can you make all the source files either `.cpp` or `.cc` (I think most are cpp already)

DSsoto:
All source files should be cpp, and all headers hpp since this is c++.

mattlangford:
We should do that in a PR soon ^

PR #17, opened 2017-04-02
PassiveSonarOS™

DSsoto:
This is a complete revamp of the passive sonar packages of the past. It is well documented, with a wiki page to come soon. The net result is a new passive sonar driver that handles everything from the acquisition of the signals via serial, the estimation of time deltas, the interface for other nodes to query the heading towards the pinger or to estimate the full position of the pinger, and the visualization of all results in RVIZ.

The code has been restructured such that all configuration can be done be the rosparam tag in a launch and some yaml. 

Also the package will work without modifications to source code for an arbitrary number of receivers in an arbitrary arrangement, as long as there is enough information to solve the multilateration problem. 

New functionality was added to the multilateration python pkg, most prominently functions to calculate the least-squares intersection of a set of (2D or 3D) lines.

The great_merge_purgatory in mil_vision was removed as promised. CameraLidarTransformer was kept, while the other files were removed.

This will all be tested at the pool on April 3rd, so we should be confident merging it after that.


sentree:
What were the results of the testing for this?

DSsoto:
All the initialization, comms code works as does visualization, multilateration and LS line intersection (including pool fixes). However I messed something up with the calculation of time deltas. Currently working on it.

drivers/mil_passive_sonar/mil_passive_sonar/passive_sonar_driver.py
```diff
@@ -0,0 +1,315 @@
...
+        rospy.init_node("passive_sonar")
+
+        self.load_params()
+        # TODO: update to use ros_alarms
+
+        try:
+            self.ser = serial.Serial(port=self.port, baudrate=self.baud, timeout=self.request_timeout)
+            self.ser.flushInput()        
+        except Exception, e:
+            print "\x1b[31mSonar serial connection error:\n\t", e, "\x1b[0m"
```
mattlangford: Use ros logging here

drivers/mil_passive_sonar/mil_passive_sonar/passive_sonar_driver.py
```diff
@@ -0,0 +1,315 @@
...
+            return None
+
+        self.reset_position_estimate(None)
+
+        self.tf_listener = tf.TransformListener()
+        self.multilaterator = Multilaterator(self.receiver_locations, self.c, self.method)
+
+        self.rviz_pub = rospy.Publisher("/visualization_marker", Marker, queue_size=10)
```
mattlangford: Consider a more unique name for vidualizations

drivers/mil_passive_sonar/mil_passive_sonar/passive_sonar_driver.py
```diff
@@ -0,0 +1,315 @@
...
+        self.receiver_count = len(locs) + 1
+        self.receiver_locations = np.zeros((self.receiver_count - 1, 3), dtype=float)
+        for i in range(self.receiver_count - 1):
+            self.receiver_locations[i] = np.array([locs[i]['x'], locs[i]['y'], locs[i]['z']])
+
+        self.method = load('method')
+        self.c = load('c')
+        self.port = load('port')
+        self.baud = load('baud')
+	self.TX_REQUEST = load('tx_request_code')
```
mattlangford: Does this work with the broken spacing?

drivers/mil_passive_sonar/mil_passive_sonar/passive_sonar_driver.py
```diff
@@ -0,0 +1,315 @@
...
+        header = make_header(frame=self.locating_frame)
+        heading = {'x' : heading[0], 'y' : heading[1],  'z' : heading[2]}
+        return {'header' : header, 'heading' : heading}
+
+    def estimate_pinger_position(self, req):
+        '''
+        Uses a buffer of prior observations to estimate the position of the pinger as the intersection
+        of a set of 3d lines in the least-squares sense.
+        '''
+        assert len(self.heading_start) > 1
```
mattlangford: Will this assertion cause the script to crash? If so it'd probably be favorable to just return a blank response with 0 observations

drivers/mil_passive_sonar/mil_passive_sonar/passive_sonar_driver.py
```diff
@@ -0,0 +1,315 @@
...
+        self.sample_variances = np.empty((0, 1), float)
+        self.pinger_position = np.array([np.NaN, np.NaN, np.NaN])
+        return {}
+
+    def reset_position_estimate(self, req):
+        '''
+        Clears all the heading and amplitude buffers and makes the position estimate NaN
+        '''
+	self.heading_start = np.empty((0, 3), float)
+	self.heading_end = np.empty((0, 3), float)
```
mattlangford: There's more spacing issues here. I know python will break under certain conditions if the spacing is off by too much.

utils/mil_msgs/srv/GetPulseHeading.srv
```diff
@@ -0,0 +1,7 @@
+# Request to receive heading to pinger
+
+---
+
+# Heading unit vector in body frame
+Header header
+geometry_msgs/Vector3 heading
```
mattlangford: Might be good to add a `is_valid` flag here

mattlangford (commented):
I might have some more comments when I'm not on mobile. Let me know if those spacing issues actually cause problems.

Will leave this as a comment pending a computer read though

DSsoto:
@mattlangford I'd hold off on more thorough reviewing until later tomorrow, there are a decent amount of changes I'll be pushing to this tomorrow, including some from pool day.

DSsoto:
pr17.re_review(reviewer = @mattlangford)

Sorry for the delay, umm.... feature creep.

drivers/mil_passive_sonar/multilateration/multilateration.py
```diff
@@ -1,73 +1,184 @@
...
+        first_times_a = (-b+math.copysign(math.sqrt(discriminant), -b))/2
+        return [first_times_a/a, c/first_times_a]
+    else:
+        return []
+
+def ls_line_intersection3d(start, end):
+    '''
+    Find the intersection of lines in the least-squares sense.
+    start - Nx3 numpy array of start points
+    end - Nx3 numpy array of end points
```
mattlangford, line 26: Could you explain this process or link to an explanation here

drivers/mil_passive_sonar/multilateration/multilateration.py
```diff
@@ -1,73 +1,184 @@
...
+        self.pairs = list(combinations(range(self.n), 2))
         self.c = c
         self.method = method
+        self.solvers = {'bancroft' : self.estimate_pos_bancroft,
+                        'LS'       : lambda dtoa: self.estimate_pos_LS(dtoa, self.cost_LS),
+                        'LS1'      : lambda dtoa: self.estimate_pos_LS1(dtoa, self.cost_LS)}
         print "\x1b[32mSpeed of Sound (c):", self.c, "millimeter/microsecond\x1b[0m"
 
-    def getPulseLocation(self, timestamps, method=None):
+    def getPulseLocation(self, dtoa, method=None):
```
mattlangford: Should_be_consistent_with_function_naming_scheme
Also should this always return something, you could get keyboard interrupted and not return anything (though anytime you get keyboard interrupted you probably want to stop the program anyway)

drivers/mil_passive_sonar/multilateration/multilateration.py
```diff
@@ -106,94 +217,96 @@ def get_B(delta):
...
+        t = self.dtoa
         x = potential_pulse[0]
         y = potential_pulse[1]
         z = potential_pulse[2]
-        d0 = np.sqrt((x0 - x)**2 + (y0 - x)**2 + (z0 - x)**2)
-        for i in xrange(1, len(self.hydrophone_locations)):
-            xi = self.hydrophone_locations[i][0]
-            yi = self.hydrophone_locations[i][1]
-            zi = self.hydrophone_locations[i][2]
+        d0 = np.sqrt((x)**2 + (x)**2 + (x)**2)
```
mattlangford: Should this be `x`, `y`, and `z` instead of `x`, `x`, `x`

drivers/mil_passive_sonar/multilateration/multilateration.py
```diff
@@ -106,94 +217,96 @@ def get_B(delta):
...
+        for i in range(self.n - 1):
+            xi = self.receiver_locations[i, 0]
+            yi = self.receiver_locations[i, 1]
+            zi = self.receiver_locations[i, 2]
             di = np.sqrt((xi - x)**2 + (yi - x)**2 + (zi - x)**2)
```
mattlangford: Same with the `x`, `y`, `z`? If so this is the same as `np.linalg.norm` which would be more faster

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+from geometry_msgs.msg import Point
+from std_msgs.msg import Header
+from mil_passive_sonar.srv import *
+
+__author__ = 'David Soto'
+
+# GLOBALS
+lock = threading.Lock()  # prevent multiple threads simultaneously using a serial port
+
+def thread_lock(lock):  # decorator
```
mattlangford, line 28: There's a `mil_tool` for this in `threading_helpers`. I think it does the same thing but it would reduce duplicated code.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+    Use an existing thread lock prevent a function from being executed by multiple threads at once
+    '''
+    def lock_thread(function_to_lock):
+        def locked_function(*args, **kwargs):
+            with lock:
+                result = function_to_lock(*args, **kwargs)
+        return locked_function
+    return lock_thread
+
+class ReceiverArrayInterface(object):
```
mattlangford, line 39: Good idea for playing back from bags vs reading from serial

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+    It is used when the keyword arg input_mode='serial' is passed to the passive sonar
+    driver constructor.
+    '''
+    def __init__(self, param_names, num_receivers):
+        self.num_receivers = num_receivers
+
+        load = lambda prop: setattr(self, prop, rospy.get_param('passive_sonar/' + prop))
+        [load(x) for x in param_names]
+
+        self.tf2_buf = tf2_ros.Buffer()
```
mattlangford, line 96: The future

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        '''
+        Request a set of digital signals from a serial port
+
+        serial_port - open instance of serial.Serial
+        data_request_code - char or string to sent to receiver board to signal a data request
+        tx_start_code - char or string that the receiver board will send  us to signal the
+            start of a transmission
+        '''
+        self.ser.flushInput()
+        try:
```
mattlangford: Wouldn't you want errors here to propagate up to the try statement in `intput_request` so that it won't do calculations with bad data?

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        '''
+        self.signals = np.full((self.num_receivers, self.signal_size), self.signal_bias, dtype=float)
+
+        for channel in range(self.num_receivers):
+            for i in range(self.signal_size):
+                while self.ser.inWaiting() < self.scalar_size:
+                    pass
+                self.signals[channel, i] = float(self.ser.read(self.scalar_size)) - self.signal_bias
+
+    def getReceiverPose(time, receiver_array_frame, locating_frame):
```
mattlangford: ~~camal case~~

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        scalar_size - size of each scalar in bytes
+        signal_bias - value to be subtracted from each read scalar before packing into array
+
+        returns: 2D numpy array with <num_signals> rows and <signal_size> columns
+        '''
+        self.signals = np.full((self.num_receivers, self.signal_size), self.signal_bias, dtype=float)
+
+        for channel in range(self.num_receivers):
+            for i in range(self.signal_size):
+                while self.ser.inWaiting() < self.scalar_size:
```
mattlangford, line 160: Should probably add some ros protection here, and maybe sleep in the loop instead of passing.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        Receives a set of 1D signals from a serial port and packs them into a numpy array
+
+        serial_port - port of type serial.Serial from which to read
+        signal_size - number of sacalar elements in each 1D signal
+        scalar_size - size of each scalar in bytes
+        signal_bias - value to be subtracted from each read scalar before packing into array
+
+        returns: 2D numpy array with <num_signals> rows and <signal_size> columns
+        '''
+        self.signals = np.full((self.num_receivers, self.signal_size), self.signal_bias, dtype=float)
```
mattlangford, line 156: This would be a good place for [a custom np.dtype](https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html).

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+
+        for channel in range(self.num_receivers):
+            for i in range(self.signal_size):
+                while self.ser.inWaiting() < self.scalar_size:
+                    pass
+                self.signals[channel, i] = float(self.ser.read(self.scalar_size)) - self.signal_bias
```
mattlangford: Does the board send 32 bit floats or 64 bit floats, I think python does everything in 64? It seems like this could cause confusion or problems at some point. It might be good to use `np.float64(...)` to make it clear what kind of data it's actually reading.

Also wouldn't it be good to do some CRC checking?

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+
+        Returns a 3x1 translation and a 3x3 rotation matrix (numpy arrays)
+        '''
+        try:
+            tfl = tf2_ros.TransformListener(self.tf2_buf)
+            T = self.tf2_buf.lookup_transform(receiver_array_frame, locating_frame, time,
+                                              timeout=rospy.Duration(0.20))
+            q = T.transform.rotation
+            t = T.transform.translation
+            rot = transformations.quaternion_matrix([q.w, q.x, q.y, q.z])
```
mattlangford, line 179: Idea for a helper function: tf(2) response to numpy position, rotation (3x3)

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+
+    For more information, read the Passive Sonar wiki page of the mil_common repository.
+    '''
+    def __init__(self, input_mode='serial', custom_input_source=None):
+
+        self.input_mode = input_mode
+        self.load_params()
+        self.bagging = False
+
+        # TODO: update to use ros_alarms
```
mattlangford, line 262: That's right

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+
+    # Passive Sonar Helpers
+
+    def load_params(self):
+        '''
+        Loads all the parameters needed for receiving and processing signals from the passive
+        sonar board and calculating headings towards an active pinger.
+
+        These parameters are descrived in detail in the Passive Sonar page of the mil_common wiki.
+        TODO: copy url here
```
mattlangford, line 296: ^

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        }
+
+        load = lambda prop: setattr(self, prop, rospy.get_param('passive_sonar/' + prop))
+        try:
+            [load(x) for x in self.required_params]
+        except KeyError as e:
+            raise IOError('A required rosparam was not declared: ' + str(e))
+
+        self.receiver_count = len(self.receiver_locations) + 1
+        self.receiver_locations = np.array(  # dictionary to numpy array
```
mattlangford, line 318: I think with custom dtypes you wouldn't have to do this, you could save the dictionary as an indexable numpy array where `self.receiver_locations['x']` gives you all the x values just like `self.receiever_locations[:, 0]` would. Food for thought

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+         'bag'    : ['bag_filename' ,'locating_frame', 'receiver_array_frame']
+        }
+
+        load = lambda prop: setattr(self, prop, rospy.get_param('passive_sonar/' + prop))
+        try:
+            [load(x) for x in self.required_params]
+        except KeyError as e:
+            raise IOError('A required rosparam was not declared: ' + str(e))
+
+        self.receiver_count = len(self.receiver_locations) + 1
```
mattlangford, line 317: Wouldn't having to specify all `receiver_locations` make more sense, does one of the hydrophones _have_ to be the origin? (Assuming the +1 is because you don't have to specify the (0, 0, 0) center hydrophone which may be bad assumption)

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+       '''
+       services = {
+                      'get_pulse_heading'        : GetPulseHeading,
+                      'estimate_pinger_position' : EstimatePingerPosition,
+                      'reset_position_estimate'  : ResetPositionEstimate,
+                      'start_bagging'            : StartBagging,
+                      'save_bag'                 : SaveBag,
+                      'set_frequency'            : SetFrequency
+                  }
+       [rospy.Service('passive_sonar/' + s[0], s[1], getattr(self, s[0])) for s in services.items()]
```
mattlangford, line 333: Do the services stay alive when you don't assign them to a variable? I feel like I've had issues where the service wouldn't be available in the past when I don't set the instantiation to a variable - thought I could have been a fool in those days.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        Heading will be a unit vector in hydrophone_array frame
+        '''
+        success, err_str = True, ''
+        signals, p0, R = None, None, None
+        try:
+            time = rospy.Time.now()
+
+            # Poll input source until it reports it is ready
+            self.input_source.input_request()
+            while not self.input_source.ready:
```
mattlangford: `and not rospy.is_shutdown():`
Even though you time out below, I feel like it'd be safer to have the control-c option in case of anomaly.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+                    raise IOError('Timed out waiting for input source')
+                else:
+                    rospy.sleep(0.05)
+            self.input_source.reset()
+
+            # Gather input from source
+            signals, p0, R = self.input_source.get_input()
+
+            # Carry out multilateration to get heading to pinger
+            heading = self.multilaterator.getPulseLocation(self.get_dtoa(signals))
```
mattlangford: This could be all zeros and throw an error, but it wouldn't be obvious why the pulse location wasn't valid when that divide by 0 error get's thrown. It would probably be good to add error logs in the `Multilaterator` class so you know why the estimate isn't valid.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+        self.visualize_pinger_pos_estimate()
+        return {'header' : {'stamp' : ros.Time.now(), 'frame_id' : self.locating_frame},
+                'num_headings' : len(self.heading_start),
+                'x' : p[0], 'y' : p[1], 'z' : p[2]}
+
+    def reset_position_estimate(self, req):
+        '''
+        Clears all the heading and amplitude buffers and makes the position estimate NaN
+        '''
+	self.heading_start = np.empty((0, 3), float)
```
mattlangford: Spacing troubles here, this doesn't cause crashing?

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+	self.heading_end = np.empty((0, 3), float)
+        self.observation_variances = np.empty((0, 0), float)
+        self.pinger_position = np.array([np.NaN, np.NaN, np.NaN])
+        return {}
+
+    def start_bagging(self, req):
```
mattlangford: I would tend to think that this shouldn't be called bagging since it's not a rosbag and can cause confusion. Maybe call it data logging or something?

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+            transparency of the marker
+        '''
+        marker = Marker()
+        marker.ns = "passive_sonar-{}".format(self.target_frequency)
+        marker.header.stamp = rospy.Time(0)
+        marker.header.frame_id = self.locating_frame
+        marker.type = marker.SPHERE
+        marker.action = marker.ADD
+        marker.scale.x = 0.2
+        np.clip(bgra, 0.0, 1.0)
```
mattlangford: This line doesn't mutate the `bgra` values, does it? `bgra = np.clip(bgra, ...)`

drivers/mil_passive_sonar/srv/EstimatePingerPosition.srv
```diff
@@ -0,0 +1,13 @@
+# Request to estimate the position of a pinger using passive sonar
+
+---
+# Response
+Header header
+
+# Number of heading observations used to calculate position estimate
+int32  num_headings
```
mattlangford: `uint32`, gotta save room for those other 2147483647+/-1 obersvations

mattlangford (changes requested):
I want pizza

I look forward to your responses.

drivers/mil_passive_sonar/srv/EstimatePingerPosition.srv
DSsoto: Done

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Good catch

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Ok, called it logging.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: I copied a tab by accident. Fixed.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: A heading will never be all zeros, but valid point about error logging in that class I'll check up on it.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Sure

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: You were a fool in those days.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: fixed

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: That's a good idea, I explicitly didn't do that or use other helpers to avoid the dependency on other mil pkgs.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Eyyyyy....

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: I need to create it first lol

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: All of this is based around the idea that there is a special receiver that all other time delta's will be measured with respect to. Therefore it should also be the basis for the coordinate system of the array. I think this is pretty reasonable and simplifies things. (You are right about the +1)

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: It sends 5 decimal ascii characters. I added some validity checking. CRC would be nice but that's up to Jake. For other users they can pass in their own object that decodes the serial bytes.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Why do you think so? I'm just converting an ascii string  to a float, and all other code here only cares about the float representation.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: I know but if we contribute this package we wouldn't want it to depend on `mil_tools`.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Haha thanks  :sunglasses:
Also going to add a Gazebo interface

drivers/mil_passive_sonar/multilateration/multilateration.py
DSsoto: Yeah this is code from my pleb tier python days. I need to fix it up and add some unit tests.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Yeah, you're right.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: I'll have to use those in the future, that sounds pretty useful. However, since this a one-off thing, I'll leave it as is.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Although it would create a dependence on another mil pkg...

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Added sleeping. What kind of ros protection? I don't see the necessity here.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
mattlangford: The dtype is composed of floats. You'd have a dtype that looks like this: `[('x', float), ('y', float), ('z', float), ('t', float)]`. Creating it would be `self.signals = np.zeros(self.num_receivers, dtype=my_dtype)`. Then filling it in would be almost the same. But using the data would make more sense. Instead of just indices`self.signals[0][0] * self.signals[1][2] + ...` you can reference fields by name `self.signals[0]['x'] * self.signals[1]['z'] + ...`. It's something I've started doing as to lessen the number of random indexing I'm doing and make it more clear to the reader of the code what's going on.

drivers/mil_passive_sonar/multilateration/multilateration.py
DSsoto: Yes

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
mattlangford: That seems strange, why not full precision floating point numbers?

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
mattlangford: Make sure to change it everywhere

drivers/mil_passive_sonar/multilateration/multilateration.py
DSsoto: Fixed, and always return, and better to throw exception than assert.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: Yeah it's a bit strange, its a Jake thing, idk.

drivers/mil_passive_sonar/multilateration/multilateration.py
```diff
@@ -1,73 +1,184 @@
 #!/usr/bin/python
 from __future__ import division
 import numpy as np
-import numpy.linalg as la
+from sklearn.preprocessing import normalize
 from scipy import optimize
 from itertools import combinations
 
 from sub8_msgs.srv import Sonar, SonarResponse
```
kev-the-dev, line 7: Should not depend on sub8, navigator

drivers/mil_passive_sonar/nodes/sonar_test.py
```diff
@@ -10,6 +10,9 @@
 import sys
 
 
+# This should be turned into a unit test
```
kev-the-dev, line 13: Should be done before merging, use rostest framework

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,590 @@
...
+
+    mode = None
+    try:
+       mode = rospy.get_param('passive_sonar/input_mode')
+    except KeyError as e:
+       mode = 'serial'
+       rospy.logerr('The param passive_sonar/input_mode was not set, defaulting to \
+                    \'serial\'. (' + str(e) + ')')
+
+    ping_ping_motherfucker = PassiveSonar(input_mode=mode)
```
kev-the-dev, line 611: We're trying to move away from potty mouth language in MIL

kev-the-dev (approved):
I'd be cool if we separate your great merge refactors from the sonar os while you're still working on it if the git operation isn't too hard

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
DSsoto: But its not even part of Carlin's list. :(

drivers/mil_passive_sonar/nodes/sonar_test.py
DSsoto: Agreed, I separated the sonar stuff from the great merge stuff and will add tests before merging.

mattlangford:
You think this is ready?

sentree:
Last time I spoke to him, @DSsoto said he was still working on ROS tests and simulation.

sentree:
Any progress on this @DSsoto?

DSsoto:
Just give me 2 More lines!

lol actually haven't really worked on this much lately, hoping to finish some of these tests by the next pool day.

mattlangford:
Ready?

sentree (changes requested):
Everything in drivers/mil_passive_sonar is currently excluded from formatting checks. Before this is merged, the check should be enabled for that directory and everything in it should be properly formatted.

drivers/mil_passive_sonar/passive_sonar_driver/passive_sonar_driver.py
```diff
@@ -0,0 +1,613 @@
...
+        print "{}\n{}".format(type(res), res)
+        return res
+
+    def estimate_pinger_position(self, req):
+        '''
+        Uses a buffer of prior observations to estimate the position of the pinger as the intersection
+        of a set of 3d lines in the least-squares sense.
+        '''
+        if len(self.heading_start) > 1:
+            raise RuntimeError(
```
kev-the-dev, line 452: I'd be better to add a success attribute to the service response and return success = False instead of crashing if this is requested without headings found

kev-the-dev (changes requested):
I encountered a scenerio when running the driver in sim in which headings were added despite being NaN. Here's a diff which fixed it:
```
diff --git a/drivers/mil_passive_sonar/passive_sonar_driver/driver.py b/drivers/mil_passive_sonar/passive_sonar_driver/driver.py
index 31942ab..b22a523 100755
--- a/drivers/mil_passive_sonar/passive_sonar_driver/driver.py
+++ b/drivers/mil_passive_sonar/passive_sonar_driver/driver.py
@@ -269,7 +269,7 @@ class PassiveSonar(object):
         try:
             variance = np.var(np.array([s.samples for s in signals]).flatten())
             # if variance > self.min_variance:  # Volume threshold
-            if np.all(np.isfinite(variance)):  # checks for NAN or inf / -inf
+            if np.all(np.isfinite(variance)) and np.all(np.isfinite(heading)):  # checks for NAN or inf / -inf
                 delta = R.dot(heading)
                 p1 = p0 + delta
 ```

kev-the-dev:
Replaced by work in mil_passive_sonar. This branch is preserved at [archive_passive_sonar](https://github.com/uf-mil/mil_common/tree/archive_passive_sonar)

DSsoto:
> Replaced by work in mil_passive_sonar. This branch is preserved at [archive_passive_sonar](https://github.com/uf-mil/mil_common/tree/archive_passive_sonar)

Oooof, almost 2 years!

PR #18, opened 2017-04-06
BATCAVE_OGRID: Publish plainer ogrid and a pointcloud

RustyBamboo:
Needs some cleaning, but it works.
- Should this be a new package? We may want to add pointcloud analysis here, or some thresholds.


RustyBamboo:
Going to make this a new package in Subjugator

PR #19, opened 2017-04-07
DEPENDENCIES: Update ros_alarms submodule to 7e3a2a6d8b94894f40bc1d5a…

kev-the-dev:
…1fe24b803eafb386


sentree (approved):
Since all of this code was reviewed in ros_alarms, this update does not need to be reviewed. Submodules allow us to version control which commit on each repository we pull. We caught this due to the CI running unit tests, so the system is already working.

PR #23, opened 2017-04-09
BLUEVIEW DRIVER: Fix bug with device connection

kev-the-dev:
tested on sub. Fixes small bug introduced when checking that we haven't gone passed all pings in blueview file.


mattlangford (approved):
Cool

PR #25, opened 2017-04-10
Remove Float64stamped

DSsoto:
Removing the msg in mil_msgs called Float64Stamped. I created two new messages to replace it, DepthStamped and RangeStamped since it was only used for depth driver and DVL driver related things. The only other usage was in the start gate mission by @RustyBamboo, where it should be removed.

@ironmig make sure you use these as you remove the dependencies on uf_common in uf-mil/odometry_tools#3 and elsewhere.

There is no order for merging these PRs that will not break our current build, I think we might as well merge this without waiting for the other things to be completed.

Also added a section to the wiki about not using msgs or srvs void of semantic meaning.

### DIFF
--- a/utils/mil_msgs/CMakeLists.txt
+++ b/utils/mil_msgs/CMakeLists.txt
@@ -22,7 +22,8 @@ add_message_files(FILES
   PoseTwistStamped.msg
   PoseTwist.msg
   VelocityMeasurements.msg
-  Float64Stamped.msg
+  DepthStamped.msg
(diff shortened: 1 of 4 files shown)


kev-the-dev (approved):
Okay, I will update sub8 integration branch once this is merged

sentree:
This built on the CI and passed all unit tests except the ros_alarms test that always fails.
-- Your friendly temporary CI guy

PR #32, opened 2017-04-13
MESSAGES: add message files for upcomming LabelMe ros interface

kev-the-dev:


utils/mil_msgs/CMakeLists.txt
```diff
@@ -22,15 +23,17 @@ add_message_files(FILES
   DepthStamped.msg
   RangeStamped.msg
   VelocityMeasurement.msg
-  Acceleration.msg
+  Acceleration.msg # Needs to be deprecated
```
DSsoto: Any chance you could get rid of Acceleration in favor of geometry_msgs/accel?

DSsoto (approved):
Would be nice... If not, will merge

kev-the-dev:
Git grep doesn't show anywhere the Acceleration message is being used in mil_common or Subjugator. It might cause problems w/ Navigator

utils/mil_msgs/msg/LabeledImage.msg
```diff
@@ -0,0 +1,7 @@
+# Represents the complete data from a labeled image from LabelMe
+
+sensor_msgs/Image image            # Image where objects are labeled
```
mattlangford: Would this be nessicary? What about a timestamp and topic to get the image from? This would greatly increase the size of bags.

utils/mil_msgs/msg/LabeledObject.msg
```diff
@@ -0,0 +1,11 @@
+# Represents a single object labeled from LabelMe
+
+string name                   # Name/Type of object entered into labeling program
```
mattlangford, line 3: Should we do an enum here to make it clear what we expect to see

utils/mil_msgs/msg/LabeledObject.msg
```diff
@@ -0,0 +1,11 @@
...
+
+string attributes             # text entered into attribute box of labeling program
+
+geometry_msgs/Point[] polygon # vertices of polygon drawn in labeling program
+                              # X,Y represent pixel cordinates in labeled image
+                              # Z is ignored, should be 0
```
mattlangford, line 9: What about a bounding box in here too? It'd probably be best to compute all bounding boxes during injestion so we don't have to worry about it later.

mattlangford (approved):
What is life without art

utils/mil_msgs/msg/LabeledObject.msg
kev-the-dev: Since this is in mil_common, no. Each project should have a wiki page / script on the usage of this for their project.

utils/mil_msgs/msg/LabeledImage.msg
kev-the-dev: The idea is that a new bag will be produced with just these messages, so all original images are just in these messages.

utils/mil_msgs/msg/LabeledObject.msg
kev-the-dev: I'm more interested in just modeling the raw data from LabelMe in ros with this. It isn't clear that all or even most of the tools using this will make use of a bounding rect.

utils/mil_msgs/msg/LabeledImage.msg
mattlangford: But when you play back to test don't you want to be able to have all original data at your disposal? Particularly Odom and anything else in the bag.

kev-the-dev: Yeah I'm starting to agree with you... maybe this should be renamed LabeledObjects.msg and just contain a header, where the stamp is the same as when the image is produced and the frame_id is the frame_id of the camera. Than you just throw this in with the bag and the tools on the other ends sync them up.

kev-the-dev: One problem with this is it is ambiguous what camera topic the labels are for. Is this for the rectified image or raw, etc.

PR #33, opened 2017-04-13
Cleanup

DSsoto:
Just improving the juju of our codebase.


perception/mil_vision/CMakeLists.txt
```diff
@@ -1,34 +1,34 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(mil_vision)
 
-# set c++11 as default, overide with set_target_properties()
+# set c++11 as default, overide with target_vompile_options()
```
kev-the-dev: Gatta fix those commented out bugs

perception/mil_vision/exFAST_SparseStereo/src/example/example.cpp
```diff
@@ -1,144 +0,0 @@
-/*
```
kev-the-dev, line 1: Why are we removing this? No chance of every using it?

kev-the-dev (changes requested):
Will approve when I see it in CI. Note, should be merged simultaneously with uf-mil/SubjuGator#218 . Good stuff

perception/mil_vision/exFAST_SparseStereo/src/example/example.cpp
DSsoto: I didn't like its performance, will be replacing with a better alternative.

perception/mil_vision/CMakeLists.txt
DSsoto: Fixed

drivers/mil_blueview_driver/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-project(blueview_driver)
+project(mil_blueview_driver)
```
sentree, line 3: @ironmig broke naming convention - sad.

utils/mil_tools/CMakeLists.txt
```diff
@@ -8,20 +8,19 @@ find_package(catkin REQUIRED COMPONENTS
   rospy
   roscpp
   cv_bridge
-  cmake_modules
 )
 
-find_package(Eigen 3 REQUIRED)
+find_package(Eigen3 REQUIRED)
```
sentree, line 13: Fixing this everywhere is going to look incredible on builds, thanks!

utils/mil_tools/src/mil_tools/mil_tools.cpp
```diff
@@ -1,6 +1,6 @@
 #include <mil_tools/mil_tools.hpp>
-#include <mil_tools/msg_helpers.h>
-#include <mil_tools/param_helpers.h>
+#include <mil_tools/msg_helpers.hpp>
+#include <mil_tools/param_helpers.hpp>
```
sentree, line 3: Have you tried using `find` to look for these? You could be sure to get them all.

sentree:
I remember you saying that this and uf-mil/SubjuGator#218 were codependent; does this require that pull request to build? We want to keep mil_common pure, so this should not depend on that at all.

sentree (approved):
Really liking the level of OCD here. As @ironmig said, this is good to merge once it builds.

DSsoto:
@whispercoros the changes are co-dependent, mil_common is not dependent on Sub

Were you able to build them together on a clean Docker image?

sentree:
This passed build, but only ran the BlueView unit tests. I'm not sure why but it looks like the ROS alarms tests didn't even build for this and #32. Check the logs and tell me if you see anything fishy.

**UPDATE:** I figured this out - the `checkout scm` command doesn't pull submodules. I will add that step to the pipeline later tonight.

sentree:
A set of codependent repositories would mean that both depend on each other (i.e. mil_common and odometry_tools). This is just SubjuGator depending on mil_common, which is normal.

kev-the-dev (approved):

PR #37, opened 2017-04-14
GNC: Absorb the odometry_tools repository

sentree:
I decided to pull request this due to lack of input on uf-mil/odometry_tools#4. I still think we should wait for input before merging, but no longer than a week.

The odometry_tools repository currently has dependencies on mil_common, but should be compilable on it's own. Since this is a package that is common to both active robots and is not a separate package we are planning on integrating, it makes sense to simple absorb it into mil_common. I preserved all of the history from that repository and moved both packages into a `gnc` directory to match the root directory conventions of SubjuGator.


sentree:
If this is merged, we should move the odometry_tools package to uf-mil-archive.

forrestv:
Fine by me. However, note that odom_estimator is only used on the sub, so
maybe it should be in the sub repo instead (leaving only odometry_utils in
mil_common)?

gnc/odom_estimator/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+    ${EIGEN_INCLUDE_DIRS}
+)
+
+add_library(odom_estimator_nodelet src/nodelet.cpp)
+target_link_libraries(odom_estimator_nodelet ${catkin_LIBRARIES})
+add_dependencies(odom_estimator_nodelet ${catkin_EXPORTED_TARGETS})
+add_dependencies(odom_estimator_nodelet ${PROJECT_NAME}_generate_messages_cpp)
+set_target_properties(odom_estimator_nodelet PROPERTIES COMPILE_FLAGS "-O3 -std=c++0x")
+
+add_executable(test_unscented_transform src/test_unscented_transform.cpp)
```
kev-the-dev, line 71: Tests should be through rostest, using python unittest or gtest.

kev-the-dev (approved):
This package could use some work on the documentation and convention front (not even a README!), but I think it makes since to absorb this as it has mil_common dependencies. Long live the reign of mil common! May we conquer all!

sentree:
Good point, odom_estimator seems SubjuGator specific in that it would have to be changed a bit to work with other projects. I'm in favor of moving that to SUbjuGator's `gnc` directory. Any objections?

DSsoto:
Also in favor of adding it to Sub gnc

DSsoto (changes requested):
Also in favor of moving odom_estimator to SubjuGator gnc.

kev-the-dev:
if you can some how git your way into this I think it is best:

* odometry_tools/odom_estimator is put into SubjuGator gnc folder with history preserved
* odometry_tools/odom_utils is put into mil_common with history preserved
* Necessary changes to CMakelist and package.xml is made in each so it builds
* odometry_tools is archived

sentree:
Looking through the code a bit, odometry_utils depends on odom_estimator. Have a look at `odometry_utils/src/transform_odometry.cpp`, which contains the fairly clear line:

    using namespace odom_estimator;

Unless these can be easily separated, it makes sense to just keep them together.

kev-the-dev:
Yeah I say just move it to Sub8.

sentree:
@DSsoto, are you in favor of keeping this here or moving it to SubjuGator?

DSsoto:
It seems best to just add to SubjuGator and later see if it can be made generic and added to mil_common.

forrestv:
NaviGator uses odometry_tools, though... so the only current option would
be adding both to mil_common, right?

I can reverse the dependency pretty easily, letting odometry_tools be put
in mil_common and odom_estimator in SubjuGator.

DSsoto:
@forrestv if you want to do that, that would be great

forrestv:
Hmm, looking at it, it would require moving a lot of the odom_estimator code over ... I think we should just take the easy route (move both into mil_common) rather than doing that right now.

DSsoto (approved):

PR #39, opened 2017-04-16
BLUEVIEW DRIVER: add noise threshold option, optional no timer pings

kev-the-dev:
Some quick fixes, should improve performance on driver with some tunning.

### DIFF
--- a/drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
+++ b/drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
@@ -14,8 +14,11 @@ class BlueViewRosDriver
 {
 public:
   BlueViewRosDriver();
+
+  /// Loads blue view settings such as range from ROS params, see example launch file
   void initParams();
-  double period_seconds;
+
+  /// Blocks until node is shutdown, generating pings and pushing them to ros
   void run();
 private:
   ros::NodeHandle nh;
@@ -29,6 +32,8 @@ private:
   cv_bridge::CvImagePtr grayscale_img, color_img;
   mil_blueview_driver::BlueViewPingPtr ping_msg;
 
+
+  double period_seconds;
   ros::Timer timer;
   bool do_grayscale, do_color, do_raw;
   std::string frame_id;
--- a/drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
+++ b/drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
@@ -15,18 +15,19 @@ class BlueViewRosDriver
 public:
   BlueViewRosDriver();
   void initParams();
-
+  double period_seconds;
+  void run();
 private:
   ros::NodeHandle nh;
   BlueViewSonar sonar;
   void loop(const ros::TimerEvent&);
+  void get_ping();
 
   image_transport::ImageTransport image_transport;
   image_transport::Publisher grayscale_pub, color_pub;
   ros::Publisher raw_pub;
   cv_bridge::CvImagePtr grayscale_img, color_img;
   mil_blueview_driver::BlueViewPingPtr ping_msg;
-  size_t i = 0;
 
   ros::Timer timer;
   bool do_grayscale, do_color, do_raw;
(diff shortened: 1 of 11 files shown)


drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
```diff
@@ -15,18 +15,19 @@ class BlueViewRosDriver
 public:
   BlueViewRosDriver();
   void initParams();
-
+  double period_seconds;
```
DSsoto: This should be replaced with getter, setter funcs (if necessary), bad practice to have public data members.

drivers/mil_blueview_driver/include/blueview_wrapper.hpp
```diff
@@ -35,9 +35,11 @@ class BlueViewSonar
 
   // Return a reference to the head for configuration like setting range
   BVTSDK::Head& getHead();
+  void updateHead(); // Must call after changes to head from getHead()
```
DSsoto: You should add doc strings for these public methods, in blueview_ros_driver.hpp as well.

drivers/mil_blueview_driver/src/blueview_ros_driver.cpp
```diff
@@ -139,20 +155,23 @@ void BlueViewRosDriver::loop(const ros::TimerEvent &)
...
     sonar.getRanges(ping_msg->bearings, ping_msg->ranges, ping_msg->intensities);
     raw_pub.publish(ping_msg);
   }
 }
+void BlueViewRosDriver::loop(const ros::TimerEvent &)
+{
+  get_ping();
+}
 
 int main(int argc, char **argv)
```
DSsoto, line 168:
```
Dude, get used to using function try blocks!

int main()
try
{
  do_stuff();
}
catch(this)
{
  handle_this();
}
catch(that)
{
  handle_that();
}
```

DSsoto (changes requested):
Two simple requests.

drivers/mil_blueview_driver/src/blueview_ros_driver.cpp
kev-the-dev: Huh? This is a try catch block

kev-the-dev:
Comments addressed but you hold on just one minute. I'm going to refactor so it's more useful to the ros community by allowing you to just untar the blueview sdk from the CD that comes with the sonar it instead of relying on our special cmake package for the sdk. refer to uf-mil/installer#11

drivers/mil_blueview_driver/src/blueview_ros_driver.cpp
DSsoto: You naive youngster..... oooooiiiiee

kev-the-dev:
Note: don't merge until uf-mil/installer#12 is merged

drivers/mil_blueview_driver/include/blueview_ros_driver.hpp
```diff
@@ -29,6 +32,8 @@ class BlueViewRosDriver
   cv_bridge::CvImagePtr grayscale_img, color_img;
   mil_blueview_driver::BlueViewPingPtr ping_msg;
 
+
+  double period_seconds;
```
DSsoto: Private data members should have a trailing underscore, according to the [ros style guide](http://wiki.ros.org/CppStyleGuide#Member_variables), especially since you want to contribute this package. this is why auto-formatting will be great.

drivers/mil_blueview_driver/src/blueview_ros_driver.cpp
```diff
@@ -167,9 +167,9 @@ void BlueViewRosDriver::loop(const ros::TimerEvent &)
...
 {
+  ros::init(argc, argv, "blue_view_driver");
```
DSsoto: Lol stickler comment, but I feel like we've been referencing blueview as one word. Consistency is always nice.

DSsoto (approved):

PR #40, opened 2017-04-16
CI: fail Jenkins build if tests dont pass

kev-the-dev:
* Now runs catkin_test_results at end of test stage of build. If any tests fail, this will return exit code 1 and therefore the build will be marked as failed. It will also display a summary in the log of what test(s) failed.


sentree (approved):
This works well on SubjuGator, so it's good to merge here.

PR #41, opened 2017-04-16
Terminal input utility, Imaging sonar launch fix

DSsoto:
* Adds utility that takes in a single character from the terminal without printing anything. (useful for programs with keyboard control, used in thruster_spin_test node in SubjuGator)
* Fixes the type used in the blueview driver example launch. (Check CMakeLists target)

### DIFF
--- a/drivers/mil_blueview_driver/launch/example.launch
+++ b/drivers/mil_blueview_driver/launch/example.launch
@@ -1,5 +1,5 @@
 <launch>
-    <node pkg="mil_blueview_driver" type="mil_blueview_driver" name="test_blue_view" >
+    <node pkg="mil_blueview_driver" type="blueview_driver" name="test_blue_view" >
         <rosparam>
             frame_id: my_sonar
             
--- a/utils/mil_tools/mil_misc_tools/terminal_input.py
+++ b/utils/mil_tools/mil_misc_tools/terminal_input.py
@@ -1,0 +1,28 @@
+#!/usr/env python
+import sys
+import tty
+import termios
+
+__author__ = 'David Soto'
+
+# Adapted from http://stackoverflow.com/questions/510357/python-read-a-single-character-from-the-user/21659588#21659588
+def get_ch():
+    '''
+    Gets a single character from stdin. Doesn't echo to screen
+    Catches CTRL-C and CTRL-E and raises exceptions.
+    '''
+    fd = sys.stdin.fileno()
+    old_settings = termios.tcgetattr(fd)
+    try:
+        tty.setraw(fd)
+        ch = sys.stdin.read(1)
+    finally:
+        termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
+
+    # Handle signal keys
+    if ch == '\x03': # CTRL-C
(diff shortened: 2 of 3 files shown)


drivers/mil_blueview_driver/launch/example.launch
```diff
@@ -1,5 +1,5 @@
 <launch>
-    <node pkg="mil_blueview_driver" type="mil_blueview_driver" name="test_blue_view" >
+    <node pkg="mil_blueview_driver" type="blueview_driver" name="test_blue_view" >
```
sentree, line 2: I thought we had agreed to move to using the mil_* prefix in this repository.

utils/mil_tools/mil_misc_tools/terminal_input.py
```diff
@@ -0,0 +1,28 @@
...
+    old_settings = termios.tcgetattr(fd)
+    try:
+        tty.setraw(fd)
+        ch = sys.stdin.read(1)
+    finally:
+        termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
+
+    # Handle signal keys
+    if ch == '\x03': # CTRL-C
+        raise KeyboardInterrupt
```
sentree, line 24: Thank you.

utils/mil_tools/mil_misc_tools/terminal_input.py
```diff
@@ -0,0 +1,28 @@
...
+    Gets a single character from stdin. Doesn't echo to screen
+    Catches CTRL-C and CTRL-E and raises exceptions.
+    '''
+    fd = sys.stdin.fileno()
+    old_settings = termios.tcgetattr(fd)
+    try:
+        tty.setraw(fd)
+        ch = sys.stdin.read(1)
+    finally:
+        termios.tcsetattr(fd, termios.TCSADRAIN, old_settings)
```
sentree, line 20: I'm really surprised that Python doesn't have a simpler way to do this, but your Stackoverflow source checks out.

sentree (approved):
You're doing this all wrong man...

This would be `terminal_input.sh`:

    #!/bin/bash
    read key
    export key

This would be `terminal_input.py`:

    #!/usr/env python
    import os
    import subprocess
    p = subprocess.Popen(terminal_input.sh , shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    p.wait()
    key = os.environ.get('key')

drivers/mil_blueview_driver/launch/example.launch
DSsoto: We did but in the cmake the target name of the executable is still blueview_driver.

mattlangford (commented):
Would just printing a few `\b` after the user types a character not delete it?

DSsoto:
@mattlangford, this is also nicer because you don't need to press enter after each key stroke to get input

drivers/mil_blueview_driver/launch/example.launch
whispercoros: I will see if I can fix this today.

PR #42, opened 2017-04-17
SCRIPTS: Cleanup and fulfill feature requests

sentree:
I went slightly overboard cleaning up the syntax used in all of our scripts, but you guys know how I do business by now. This is the syntax I've standardized on:

* `if [[ <logical_comparison> ]]; then`
* `if (( <numerical_comparison> )); then`
* Checking strings for equality to `true` or `false` for booleans
* Using `==` as opposed to `=` in logical comparisons
* Using `$()` as opposed to `` for in-lining commands

A few of the features requested last testing day were also implemented:

* The `wsmux` command now changes the `CATKIN_DIR` environment variable of a shell
* The `-n` argument of `ros_connect` is now autocompleted with all known hostnames
* The `bmux` command clears before sourcing new bagging variables

Support for Git submodules was also added to the build pipeline in the Jenkins file.

### DIFF
--- a/scripts/bash_aliases.sh
+++ b/scripts/bash_aliases.sh
@@ -24,13 +24,13 @@ ros_env() {
 
 
 # Directory navigation
-alias mc="cd $CATKIN_DIR/src/mil_common"
+alias mc="cd \$CATKIN_DIR/src/mil_common"
 
 # Bash sourcing
 alias srcbrc="source ~/.bashrc"
 
 # Catkin workspace management
-alias cm="catkin_make -C $CATKIN_DIR -j8"
+alias cm="catkin_make -C \$CATKIN_DIR -j8"
 
 # Alarms
 alias araise="rosrun ros_alarms raise"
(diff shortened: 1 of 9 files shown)


scripts/bash_aliases.sh
```diff
@@ -24,13 +24,13 @@ ros_env() {
 
 
 # Directory navigation
-alias mc="cd $CATKIN_DIR/src/mil_common"
+alias mc="cd \$CATKIN_DIR/src/mil_common"
```
kev-the-dev, line 27: What does this do?

kev-the-dev (commented):
Appreciate the clean ups. Bash is now somewhat more sane. Since we don't have tests for these scripts, have you tested these changes?

scripts/bash_aliases.sh
sentree: The change escapes the `$` character, which means the variable's value will be based on when the alias is used rather than when it is set.

sentree:
I have tested them heavily and we used them at the last test day.

DSsoto (approved):

PR #43, opened 2017-04-18
BUG: Fix eigen include on some systems

kingkevlar:
Debian 8.7


mattlangford (commented):
I've never seen this issue before, what is the root cause?

sentree:
Let's try to keep this discussion on uf-mil/SubjuGator#225 since the pull requests are so similar.

kev-the-dev (approved):
Good. Should prevent some edge case build errors.

DSsoto (approved):
Approve. The issue was the capitalization of eigen. Apparently on Debian there are multiple eigen_include_dirs variables and the capitalization affected which include paths were added.

PR #44, opened 2017-04-22
LABELING: add script to interface between LabelMe and ROS bags

kev-the-dev:
Add a script which takes a YAML configuration and can generate ROS bags with labeled image data or put images for ROS bags into a LabelMe installation. Can also generate a report on what percentage of images have been labeled. See argparse config at bottom for usage.

Has been tested on a LabelMe installation successfully with provided example config. From this yaml, images were put onto LabelMe. Once labeled, the same script was use to produce a new bag with all of the original data with label data added at correct times.

<small>#Langforded</small>


utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env python
+
+"""
+Converts bags to labelme images, or labelme annotations to bags.
+
+Uses a YAML file to specify what times and topics in a set of bag files
+will be used to produce images which are places in a corosponding
```
sentree: Looks like you got the Langford brand of spelling down as well.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,231 @@
...
+          obj_msg.id = int(obj.find('id').text)
+          parent = obj.find('parts').find('ispartof').text
+          if parent != None:
+            obj_msg.parent_id = int(parent)
+          attributes = obj.find('attributes').text
+          if attributes != None:
+              obj_msg.attributes = attributes
+          polygon = obj.find('polygon')
+          for pt in polygon.findall('pt'):
+            x = int(pt.find('x').text)
```
mattlangford: Woah cowboy, slow down. Mixing yer 4 and 2 spaces.

utils/mil_tools/nodes/label_me_bag_example.yaml
```diff
@@ -0,0 +1,14 @@
...
+    labeled_file: orange_pipes_labeled.bag
+    segments:
+      - name: path_marker_pool_1
+        start: 1.0
+        stop: 10.0
+        topic: /down/left/image_raw
+        freq: 0.5
+      - name: path_marker_pool_2
+        start: 15.0
+        stop: 20.0
```
mattlangford: Can you support -1.0 or something for a `stop:` (to dump until the end of the bag).

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,231 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generates rosbags based on LabelMe data and visa/versa')
+    parser.add_argument('config', type=str,
+                        help='YAML file specifying what bags to read and extract images from. See example YAML for details')
+    parser.add_argument('--labelme-dir', '-d', dest='dir',type=str, required=True,
+                        help='root directory of labelme instalation')
+    parser.add_argument('--extract', '-e', dest='extract_labels', action='store_true',
+                        help='Read annotations from labelme, inserting them into a new bag')
+    parser.add_argument('--report', '-r', dest='do_report', action='store_true',
```
mattlangford: If you have shorthand args like `-r` it may be good to make the full length arguments more descriptive, i.e.`--extract-labels` or `--report-labels`. It will be easier to remember when you have to call this script a month or two down the line (so you're not having to `-h` everytime you run the script). Also when you have scripts with 10 or so arguments, clear names help a lot.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,231 @@
...
+              if not 'name' in segment:
+                  raise Exception("Segment does not contain 'name' attribute {}".format(segment))
+              if not 'freq' in segment:
+                  raise Exception("Segment does not contain 'freq' attribute {}".format(segment))
+
+  def _read_bag(self, bagfile, segments):
+      bag = rosbag.Bag(bagfile)
+      _, _, first_time = bag.read_messages().next()
+      for segment in segments:
+          print "\tProccessing Segment '{}' with topic '{}'".format(segment['name'], segment['topic'])
```
mattlangford: `fprint`?

utils/mil_tools/nodes/label_me_bag_example.yaml
```diff
@@ -0,0 +1,14 @@
...
+        topic: /down/left/image_raw
+        freq: 1.0
```
mattlangford: Also, you should allow 0.0 `freq` to dump everything. (You'd have to check for zero in `_read_bag` else you'd divide by 0)

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,231 @@
...
+      total_xml_count = 0
+      total_img_count = 0
+      for bag in self.config['bags']:
+          print "\tGenerating completion report for {}".format(bag['file'])
+          x, i = self._print_bag_report(bag)
+          total_xml_count += x
+          total_img_count += i
+      print "{}/{} TOTAL images labeled ({}%)".format(total_xml_count, total_img_count, round(100.0*total_xml_count/total_img_count,2))
+
+  def _verify_yaml(self):
```
mattlangford: Maybe have this function print the intent of the script, i.e. "I will dump images from {topic} to {outdir} on {bagname} from {start} to {end} every {freq} seconds." for each entry in the yaml. I know you do this down in the bag function, but having it all print out at once may be more useful. That would let you verify everything will run smoothly and leave your computer and let it work (I don't know how long this takes to run but on multiple big bags it could be rough)

This would have been cool for the mission runner back in the good old days of Navigator.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,231 @@
...
+          xml_count = 0
+          img_count = 0
+          for xmlfile in os.listdir(xml_path):
+              xml_count += 1
+          for imgfile in os.listdir(img_path):
+              img_count += 1
+          print "\t\t\t{}/{} images labeled in this segment ({}%)".format(xml_count, img_count, round(100.0*xml_count/img_count,2))
+          total_xml_count += xml_count
+          total_img_count += img_count
+      print "\t\t{}/{} images labeled in this bag ({}%)".format(total_xml_count, total_img_count, round(100.0*total_xml_count/total_img_count,2))
```
mattlangford: You do the percent function a lot, maybe break it out somewhere to clean up code?

mattlangford (approved):
I think this is a good first run. It may be difficult to get times for every bag you want to label though right? That would require going through all the bags you want images for and keeping track of start and end times.
This might be a good case for an addition to the rqt bag plugin that would let you play a bag and make selections there.

What do you think the intended workflow would be?

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: Great idea. I'll also add a dry run flag to describe what it going to happen based on YAML and not actually touch and files.

utils/mil_tools/nodes/label_me_bag_example.yaml
kev-the-dev: Would be useful. I'll probably just accept a missing start or stop parameter and assume 0.0 and END_OF_BAG respectively.

kev-the-dev:
While it will be tedious to select segment times, I'm not sure of any other option (if a program could do it, there would be no point to our perception). For finding times for the path marker task, it wasn't too hard. I just pulled up rqt and toggled the bag with spacebar to pick some good times. Writing notes on what is in bags when we take them will help a lot.

mattlangford:
Yeah, taking bags specific for the vision task is a good idea.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+      config: path to YAML file following format of the example yaml file in this package
+      labelme_dir: root directory of labelme instalation (should contain Images and Annotations directories)
+      """
+      self.verbose = verbose
+      if not os.path.isdir(labelme_dir):
+          raise Exception("Labelme directory {} does not exsist".format(labelme_dir))
+
+      config_file = open(config, 'r')
+      self.config = yaml.load(config_file)
+      self._verify_yaml()
```
sentree: I like that you're using private functions and variables here, but consider using double underscores at the beginning of the name. A single underscore implies private for other programmers and fits our convention, but double underscores will actually make it hard to access from outside the object. Since this is more similar to a truly private function or variable, it is what I use.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+      self.bridge = CvBridge()
+      self.labelme_dir = labelme_dir
+
+  def _print(self, string,*args):
+      if self.verbose:
+          print string.format(*args)
```
sentree: I feel like python has a more elegant logging interface, but it's not a big deal.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+      """
+      Crawls through all bags in config YAML, placing images from the specified
+      topics and the specified frequency into corosponding folders within
+      Images of the labelme instalation.
+      
+      These folders will be created as follows if not already:
+      LABELME_DIR/Images/SEGEMENT_NAME/TOPIC_NAME
+      where TOPIC_NAME replaces all / with @ for compatibility with filesystems.
+      """
+      for bag in self.config['bags']:
```
sentree: Do you intend to have a master YAML for all of the bags?

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+
+  def extract_labels(self):
+      """
+      Finds labeled XML files in LabelMe directories determined from the
+      config file, placing parsed annotation data as ros messages in a new bag
+      """
+      self._print("Extracting labels for all bags in config")
+      for bag in self.config['bags']:
+          self._print("\tExtracting LabelMe annotations for {}".format(bag['file']))
+          # TODO: give output bag a default name like INFILE_labeled.bag
```
sentree: This should probably happen before merging.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+          outfile = bag['labeled_file']
+          self._extract_labels_bag(bag, outfile)
+
+  def print_report(self):
```
sentree: Can you return the percentage so this is usable by another program? I think I could have the fully annotated bags be automatically assembled by a simple cron job on the server.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+          self._print("\tGenerating completion report for {}", bag['file'])
+          x, i = self._print_bag_report(bag)
+          total_xml_count += x
+          total_img_count += i
+      if total_img_count == 0:
+          print "{}/{} TOTAL images labeled (0%)".format(total_xml_count, total_img_count)
+      else:
+          print "{}/{} TOTAL images labeled ({:.1%})".format(total_xml_count, total_img_count, total_xml_count/total_img_count)
+
+  def _read_bag(self, bagfile, segments):
```
sentree: Please add a documentation string to this function. Comments would also be nice if you have time.

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+                    labels = self._parse_label_xml(label_files[topic][str(msg.header.stamp)])
+                    labels.header = msg.header
+                    out.write(topic+'/labels', labels, t)
+        out.write(topic, msg, t)
+      out.flush()
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generates rosbags based on LabelMe data and visa/versa')
+    parser.add_argument('config', type=str,
+                        help='YAML file specifying what bags to read and extract images from. See example YAML for details')
```
sentree: Why no `--` before the option?

utils/mil_tools/nodes/label_me_bag.py
```diff
@@ -0,0 +1,268 @@
...
+    parser.add_argument('--labelme-dir', '-d', dest='dir',type=str, required=True,
+                        help='root directory of labelme instalation')
+    parser.add_argument('--extract-labels', '-e', dest='extract_labels', action='store_true',
```
sentree: I don't know if extract is the right word for this. I would associate that more with the image extraction process. Maybe use inject?

utils/mil_tools/nodes/label_me_bag_example.yaml
```diff
@@ -0,0 +1,13 @@
+bags:                                       # List of bags to generate images for labeling from
+  - file: orange_pipes.bag                  # File to pull images from
+    labeled_file: orange_pipes_labeled.bag  # Output file for original data + label topics
+    segments:                               # List of named segments in bag to get images from
+      - name: path_marker_1                 # Name of segment to appear in LabelMe
+        start: 1.0                          # Duration in seconds to start getting images, optional
+        stop: 2.0                           # Duration in seconds to stop getting images, optional
+        topic: /down/left/image_raw         # Topic to pull images from
```
sentree: What frequency will images in this time period be pulled?

sentree (changes requested):
This is a good start on a library for interfacing with LabelMe. That said, I think this should be split up a bit. There should be a library and a CLI.

I think an RQT interface would make generating the configurations much simpler for people not familiar with the system. Having a separate library would make the RQT and CLI interfaces compatible and avoid repeated code. Basically, users would use one of the tools to extract images to the fileserver and the bag would automatically be reassembled with the labels once it is done.

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: +1 for cron jobs

Right now without verbose it just returns one line where the last token is a percentage. You could use that.

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: It is a positional argument not a flag.

utils/mil_tools/nodes/label_me_bag.py
mattlangford: Not since I found `:set spell`, vims built in comment spell checker.

utils/mil_tools/nodes/label_me_bag.py
mattlangford:
```
Use one leading underscore only for non-public methods and instance variables.

To avoid name clashes with subclasses, use two leading underscores to invoke Python's name mangling rules.
```

utils/mil_tools/nodes/label_me_bag.py
sentree: Everything is actually spelled right here, they're just not the right words. I don't really know what that's called. Thanks for the tip, though.

utils/mil_tools/nodes/label_me_bag.py
sentree: Where did that come from? The only issue I have with one underscore is it really does nothing to make a function or variable private, it's just for looks.

sentree:
Would it be useful to add `-n` and `--notes` arguments to the bag command that open a text file of the same name as the bag in Vim after the bag is ended?

utils/mil_tools/nodes/label_me_bag.py
mattlangford: Oh yeah, leaving out some words I see - though `corresponding` is spelled wrong

utils/mil_tools/nodes/label_me_bag.py
mattlangford:
```
We don't use the term "private" here, since no attribute is really private in Python
```
All from the PEP8 page

utils/mil_tools/nodes/label_me_bag_example.yaml
kev-the-dev: It explains in the comment on the next line that if no freq is specified, every image from this time frame will be inserted into LabelMe

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: I'm going to make it a flag though so it's more verbose.

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: renamed to ```--inject-annotations```

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: I would prefer to have one YAML per perception task, containing all of the bags/segments used to test/train for that task. So each YAML would effectively have one maintainer, whoever is assigned that perception task. These YAMLS should be version controlled, probably in the project's git.

utils/mil_tools/nodes/label_me_bag.py
kev-the-dev: Most python libraries I look at, like rospy, use one underscore for class "private" variables/methods, so I emulated that. Python is a scripting language, so I don't think it's that important to enforce private variables when there isn't even type safety. You just got to believeeee

kev-the-dev:
Tony's comments have been addressed. I don't think we should hold this PR up because there is not graphical interface yet. And it currently can be used as a library, with the usual ``` "if __name__ == "__main__" ``` at the bottom just providing an optional argparse CLI. The class can be used normally in a python script.

whispercoros:
I think this is good to merge, but should we wait for a review from @DSsoto?

sentree (approved):
I think this is good to merge, but should we wait for a review from @DSsoto?

kev-the-dev:
I think I fixed the unrelated build error. Should be good to go

PR #45, opened 2017-04-24
CLEANUP: Organize online_bagger

kingkevlar:
I organized online_bagger to match the existing file structure in mil_common and made the necessary (one line) change to the actual file (online_bagger.py) required from moving the .srv file from mil_tools to mil_msgs (with the rest of the msgs and srvs).


utils/mil_tools/mil_ros_tools/online_bagger/config/online_bagger_example.yaml
```diff
@@ -1,15 +0,0 @@
-# list an arbitrary set of topics to  subscribe to and stream:
```
mattlangford: I think this is useful to keep as an example in an example launch.
Idea:
Move `online_bagger.py` into `utils/mil_tools/nodes`, then making a `utils/mil_tools/launch` where you put the launch file with the config in there as params in the launch.

mattlangford (approved):
A suggestion for organization, take it or leave it

utils/mil_tools/mil_ros_tools/online_bagger/config/online_bagger_example.yaml
kingkevlar: Give me two more lines

sentree (changes requested):
This will not work as it is. In order to build the message in the mil_msgs package, you need to add `BaggerCommands.srv` to the `add_service_files` block in `utils/mil_msgs/CMakeLists.txt`.

kev-the-dev (changes requested):
The example launch file now refers to a config file that isn't there. Should change before merging

sentree:
Is there a reason we're keeping the sample launch file and not the sample configuration file? Realistically, I think both of these should be in a readme with some good documentation.

    @kingkevlar --verbose

kev-the-dev (approved):

kingkevlar:
@whispercoros It should be there now (both the config and launch file)

sentree (approved):

PR #47, opened 2017-04-28
SCRIPTS: Add SSH client binding to wsmux

sentree:
This implements SSH client binding for vehicles and workstations where there are multiple catkin workspaces on a user account. It will only work for situations in which the user is connecting remotely via SSH. Running `wsmux -b [WORKSPACE]` will bind the IP address of the machine you are connecting from to a specific workspace.

Running `wsmux -c` will resolve and connect to the workspace that an SSH client is bound to. This should be added to the bash runcom files of vehicles and workstations to make it automatic.

If the script detects a `.wsrc` file in the root directory of the workspace, it will be sourced when `wsmx -c` is run, but not when just `wsmux` is run. This will allow us to have user specific configurations while maintaining the ability to switch to another users workspace temporarily.

### DIFF
--- a/scripts/wsmux.sh
+++ b/scripts/wsmux.sh
@@ -101,6 +101,11 @@ wsmux() {
 						if [[ -f ~/$(echo $BINDING | cut -d ':' -f2)/devel/setup.sh ]]; then
 							SELECTION=~/$(echo $BINDING | cut -d ':' -f2)
 
+							# Source the workspace runcom file if it exists
+							if [[ -f $SELECTION/.wsrc ]]; then
+								source $SELECTION/.wsrc
+							fi
+
 						# Otherwise, prompt the user to check themself
 						else
 							echo "The specified workspace is not valid. I don't trust like that..."
--- a/scripts/wsmux.sh
+++ b/scripts/wsmux.sh
@@ -21,7 +21,7 @@ _catkin_ws_complete() {
 		if [[ -f $FILE/$WS_MARKER ]]; then
 
 			# Append just the name of the workspace folder to the autocomplete list
-			COMPREPLY+=( $(echo "$FILE" | rev | cut -d "/" -f1 | rev) )
+			COMPREPLY+=( $(echo "$FILE" | rev | cut -d '/' -f1 | rev) )
 		fi
 	done
 }
@@ -56,7 +56,7 @@ wsmux() {
 			-s|--show)
 				if [[ ! -z "$ROS_PACKAGE_PATH" ]]; then
 					echo -n "Currently sourced catkin workspace: "
-					echo "$ROS_PACKAGE_PATH" | cut -d ":" -f1 | sed "s@/src@@"
+					echo "$ROS_PACKAGE_PATH" | cut -d ':' -f1 | sed "s@/src@@"
 				else
 					echo "No catkin workspace is currently sourced"
 				fi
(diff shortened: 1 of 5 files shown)


DSsoto (commented):
I couldn't find anything wrong here, looks cool!
Let @ironmig look at it before merging.

scripts/wsmux.sh
```diff
@@ -95,13 +159,18 @@ wsmux() {
...
+			source $SELECTION/devel/setup.sh
+			export CATKIN_DIR=$SELECTION
 	fi
 
 	unset COMPREPLY
 }
 
+# Generates the configuration file if it does not exist
+if [[ ! -f $WS_CONFIG_FILE ]]; then
+	echo "WORKSPACE_BINDINGS=(	" > $WS_CONFIG_FILE
```
kev-the-dev, line 171: Why isn't this just one line?

scripts/wsmux.sh
```diff
@@ -36,14 +39,81 @@ wsmux() {
...
+					echo "Please specify the name of a workspace in the home directory"
+
+				else
+					source $WS_CONFIG_FILE
+
+					# Replace any binding that already exists for this SSH client
+					local REPLACED="false"
+					for (( BINDING_INDEX=0; BINDING_INDEX < ${#WORKSPACE_BINDINGS[@]}; BINDING_INDEX++ )); do
+						if [[ "$(echo ${WORKSPACE_BINDINGS[BINDING_INDEX]} | cut -d ':' -f1)" == "$(echo $SSH_CLIENT | cut -d ' ' -f1)" ]]; then
+							WORKSPACE_BINDINGS[BINDING_INDEX]="$(echo $SSH_CLIENT | cut -d ' ' -f1):$2"
```
kev-the-dev, line 72: So this is based on IP address right? But won't this change every time we reconnect beacause of DHCP?

scripts/wsmux.sh
```diff
@@ -36,14 +39,81 @@ wsmux() {
...
-				echo "-s		--show			Show the selected catkin workspace"
+				echo "-l		--list			List the detected workspaces"
+				echo "-s		--show			Show the selected workspace"
+				SELECTION="false"
+				shift 1
+				;;
+			-b|--bind)
+
+				# Ensure that a client has connected via SSH
+				if [[ -z "$SSH_CLIENT" ]]; then
```
kev-the-dev, line 57: Perhaps there should be an optional  "default" or "local" line in the wsmux.conf for when not over SSH. So in our case it will source mil_ws if non on ssh or if ip address is not on config

kev-the-dev (approved):
Useful feature addition. From reading it I think it doesn't change any behavior by default, so I'm find to merge, but consider adding the local/default option.

scripts/wsmux.sh
kev-the-dev: It might be tough, but ideally it would get our hostname and bind based on that.

scripts/wsmux.sh
sentree: The MIL runcom will handle selecting the main workspace, this doesn't change that behavior.

scripts/wsmux.sh
sentree: Your DHCP leases are all handled based on the MAC addresses you gave me, so the IPs will not change.

scripts/wsmux.sh
sentree: This could be done in one line, but it would get very long as more and more users bind their sessions. I figured keeping one binding per line would be more effective formatting.

PR #48, opened 2017-04-30
ONLINE BAGGER: Changes to allow integration with the bag command

kev-the-dev:
* add much more configurability to where bags are placed from service call. Can now specify directory, will add .bag if not present, resolve relative directories, etc

* add topics argument to service. If not empty string, only space separated list topics will be bagged

* read BAG_ALWAYS and bag_* environment variables to get topics to buffer from bmux

PR #49, opened 2017-04-30
SCRIPTS: Bag OS (ɔ)

sentree:
Preface: Note the copyleft, for open source software *(Ahem, #17)*

These changes should complete work on the bagging system for the most part. They are dependent on #48 and uf-mil/SubjuGator#228. To brieefly summarize what has been done:

* The `bmux` command has been merged into the `bag` command in the form of a `-i` flag. This means that `bag -i <project_repository_name>` should replace `bmux <project_repository_name>` in the Bash runcom file on all of the vehicles.
* A `-e` flag has been added to export configurations to a file. This means that new groupings created with the `-g` flag can easily be saved.
* A `-t` flag was added to trigger a note taking mode after the bag has been saved. It will open a text file in the same directory as the bag file and with the same name in Vim. Hopefully this makes it easier to grep through the bag directories at a later date to find what we need.
* Support for the online bagger was added with the `-o` flag. Passing this along with <time> will trigger the dumping of a bag containing the desired topics that begins at t - <time>.

Everything here except the interaction with the online bagger has been tested and works well. Double check my syntax on the `rosservice` command, but I believe it is correct.


scripts/bag.sh
```diff
@@ -54,30 +115,76 @@ bag() {
...
+				echo "-c		--clear			Clear the existing configuration"
 				echo "-d [DIRECTORY]	--directory		Set the bag storage directory"
-				echo "-g 		--group			Group a set of existing variables"
+				echo "-e [REPOSITORY]	--export		Export the configuration to a file"
+				echo "-g [BAG_VARS]	--group			Group existing variables and topics"
 				echo "-h		--help			Display the help menu"
+				echo "-i [REPOSITORY]	--import		Import a configuration from a file"
 				echo "-l		--list			List available bagging variables"
 				echo "-n [BAG_NAME]	--name			Pass a name for bags or groups"
+				echo "-o [TIME]	--online		Bag a number of seconds in the past"
```
kev-the-dev, line 169: Is this going to be misaligned with the rest of the help text?

scripts/bag.sh
```diff
@@ -134,20 +252,50 @@ bag() {
...
+	# If online mode was selected, call the online bagger service
+	elif [[ "$MODE" == "online" ]]; then
+
+		# Retrieve a bag name if one was not passed
+		while [[ -z "$NAME" ]]; do
+			echo -n "What should this bag be called? " && read NAME
+		done
+
+		# Call to the online bagger with what topics to dump and where
+		rosservice call /online_bagger/dump "{bag_name: '$NAME', topics: '$TOPICS', bag_time: '$TIME'}"
```
kev-the-dev: You should pass in the full path here.
 ```rosservice call /online_bagger/dump "{bag_name: '$BAG_DIR/$(date +%Y-%m-%d)/$NAME', topics: '$TOPICS', bag_time: '$TIME'}"``` should work

scripts/bmux.sh
```diff
@@ -1,110 +0,0 @@
-#!/bin/bash
```
kev-the-dev, line 1: It is always a sad day when we lose a *mux script, but alas we must move on.

kev-the-dev (changes requested):
As #48 is implemented right now, you must pass the full directory in to the bag name to ensure it gets in the right place. Once thats done we gud

scripts/bag.sh
sentree: Nope. Don't question it, it's the tab lifestyle...

scripts/bmux.sh
sentree: I'm sure one will spring up soon to replace it.

sentree:
The points that @ironmig made have been addressed.

scripts/bag.sh
```diff
@@ -1,40 +1,101 @@
...
 
+generate_configuration() {
+	local BAGGING_VARIABLES="$(env | grep 'bag_' | cut -d '=' -f1 | cut -d ' ' -f2)"
+	local VARIABLE
+
+	echo "#!/bin/bash"
+	echo ""
+	echo "# This script will be sourced by the bag command as a configuration file. It is"
+	echo "# imperative that each variable is in the bag_* namespace and contains a space"
+	echo "# diliniated list of ROS topics. The only two exceptions to this rule are the"
```
DSsoto: deliniated

DSsoto (approved):
This is pretty good Tony, hope you will soon produce equally good mission/perception code.

sentree:
The points that @dsoto made have been addressed.

DSsoto:
Good to merge when @ironmig updates review.

kev-the-dev (approved):

PR #51, opened 2017-05-09
BUG: Fix resubscribe period on online_bagger 

kingkevlar:
-This fixes the severe slow down of online_bagger caused by a constant attempt to resubscribe to multiple topics literally every single callback as pointed out by @ironmig:
    Added a rosparam resubscribe_period to the config 
-Fix bag.sh to provide it the correct parameters required to use online_bagger (@whispercoros)
-Fix bug related to partial bagging of available topics on edge cases (when zero or only one message is available)
-Also made improvements to output for readability

### DIFF
--- a/utils/mil_tools/launch/online_bagger_example.yaml
+++ b/utils/mil_tools/launch/online_bagger_example.yaml
@@ -1,7 +1,10 @@
 # list an arbitrary set of topics to  subscribe to and stream:
+# overwrite default stream time on specific topics in this config
 
 stream_time: 10  # seconds
 
+# adds additional topics to stream, or overrides default stream time on specific topics
+# leave blank to bag MIL bag script enviromental variables only
 topics: [["/odom",                    300] ,
          ["/absodom",                 300],
          ["/stereo/left/camera_info"     ],
@@ -11,5 +14,8 @@ topics: [["/odom",                    300] ,
          ["/velodyne_points",         300],
          ["/ooka",                    100]]
 
-# comment out if default home/user/online_bagger is desired
+# determines how frequently unavailable topics are resubscribed to defaults to 3.0 seconds:
+resubscribe_period: 3.0 # seconds
+
+# comment out if default env variable 'BAG_DIR'/current_date is desired:
 # bag_package_path: ''  
--- a/utils/mil_tools/launch/online_bagger_example.yaml
+++ b/utils/mil_tools/launch/online_bagger_example.yaml
@@ -12,10 +12,15 @@ topics: [["/odom",                    300] ,
          ["/stereo/right/camera_info", 20],
          ["/stereo/right/image_raw",   20], 
          ["/velodyne_points",         300],
-         ["/ooka",                    100]]
+         ["/tf",                    100]]
 
 # determines how frequently unavailable topics are resubscribed to defaults to 3.0 seconds:
 resubscribe_period: 3.0 # seconds
 
 # comment out if default env variable 'BAG_DIR'/current_date is desired:
-# bag_package_path: ''  
+# bag_package_path: ''
+
+# Bool parameter that determines if current date is appended to bag directory
+# for example defaults to: "home/user/bags/current-date" when True
+# else bag directory goes to: "home/user/bags when False
+dated_folder: True 
--- a/utils/mil_tools/nodes/online_bagger.py
+++ b/utils/mil_tools/nodes/online_bagger.py
@@ -47,7 +47,16 @@ class OnlineBagger(object):
                                              self.start_bagging)
 
         self.subscribe_loop()
-        rospy.loginfo('subscriber list: %s', self.subscriber_list)
+
+        rospy.loginfo('')
+        rospy.loginfo('Remaining Failed Topics:')
+        self.print_failed_list()
+        rospy.loginfo('')
+
+    def print_failed_list(self):
+        for topic in self.subscriber_list.keys():
+            if self.subscriber_list[topic][1] == False:
+                rospy.loginfo('{}, {}'.format(self.subscriber_list[topic], topic))
 
     def get_params(self):
 
@@ -84,8 +93,9 @@ class OnlineBagger(object):
             if key[0:4] == 'bag_':
                 add_env_var(os.environ[key])
 
-        rospy.loginfo('subscriber list: %s', self.subscriber_list)
-        rospy.loginfo('stream_time: %s seconds', self.stream_time)
+        self.resubscribe_period = rospy.get_param('~resubscribe_period', default=3.0)
+
+        rospy.loginfo('Default stream_time: {} seconds'.format(self.stream_time))
 
     def make_dicts(self):
 
@@ -130,7 +140,9 @@ class OnlineBagger(object):
         for topic in self.subscriber_list:
             self.topic_messages[topic] = SliceableDeque(deque())
 
-        rospy.loginfo('topics status: %s', self.subscriber_list)
+        rospy.loginfo('')
+        rospy.loginfo('Initial subscriber_list:')
+        self.print_failed_list()
 
     def subscribe_loop(self):
 
@@ -148,8 +160,11 @@ class OnlineBagger(object):
             i = i + 1
             if i % 1000 == 0:
                 rospy.logdebug('still subscribing!')
+        rospy.loginfo("Subscribed to {} of {} topics, will try again every {} seconds".format(self.successful_subscription_count, 
+            len(self.subscriber_list), self.resubscribe_period))
+        rospy.Timer(rospy.Duration(self.resubscribe_period), self.subscribe)
 
-    def subscribe(self):
+    def subscribe(self, time_info=None):
 
         """
         Subscribe to the topics defined in the yaml configuration file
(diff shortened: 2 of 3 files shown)


utils/mil_tools/launch/online_bagger_example.yaml
```diff
@@ -1,7 +1,10 @@
 # list an arbitrary set of topics to  subscribe to and stream:
+# over write default stream time on specific topics in this config
```
DSsoto: Overwrite is one word

utils/mil_tools/launch/online_bagger_example.yaml
```diff
@@ -11,5 +14,8 @@ topics: [["/odom",                    300] ,
          ["/velodyne_points",         300],
          ["/ooka",                    100]]
 
-# comment out if default home/user/online_bagger is desired
-# bag_package_path: ''  
\ No newline at end of file
+# determines how frequently unavailable topics are resubscribed to:
+resubscribe_period: 3.0 # seconds
```
DSsoto: This is much better than trying to resubscribe every callback!

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -47,7 +47,16 @@ def __init__(self):
                                              self.start_bagging)
 
         self.subscribe_loop()
-        rospy.loginfo('subscriber list: %s', self.subscriber_list)
+
+        rospy.loginfo('')
```
DSsoto: This will just print out the time with no log message. I'm assuming this is not the behaviour you want. Did you mean to just put newlines before and after your messagge? If so, just do:
`rospy.loginfo('\nRemaining Failed Topics:\n%s\n', self.get_subscriber_list_string())
`

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -47,7 +47,16 @@ def __init__(self):
...
+        rospy.loginfo('Remaining Failed Topics:')
+        self.print_subscriber_list(False)
+        rospy.loginfo('')
+
+    def print_subscriber_list(self, status):
```
DSsoto: Add a docstring, and if you want to use my suggestion above, make this function not print anything, just return a string.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -47,7 +47,16 @@ def __init__(self):
...
+        for topic in self.subscriber_list.keys():
+            if self.subscriber_list[topic][1] == status:
+                rospy.loginfo('{}, {}'.format(self.subscriber_list[topic], topic))
 
     def get_params(self):
```
DSsoto: It feels like this function should have a docstring. Are there any parameters that are necessary (don't have defaults)?

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -84,9 +93,10 @@ def add_env_var(var):
             if key[0:4] == 'bag_':
                 add_env_var(os.environ[key])
 
-        rospy.loginfo('subscriber list: %s', self.subscriber_list)
-        rospy.loginfo('stream_time: %s seconds', self.stream_time)
+        self.resubscribe_period = rospy.get_param('~resubscribe_period', default=3.0)
 
+        rospy.loginfo('Default stream_time: {} seconds'.format(self.stream_time))
+        rospy.loginfo('Bag Directory: {}'.format(self.dir))
     def make_dicts(self):
```
DSsoto: Space between functions

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -130,7 +140,9 @@ def __getitem__(self, index):
         for topic in self.subscriber_list:
             self.topic_messages[topic] = SliceableDeque(deque())
 
-        rospy.loginfo('topics status: %s', self.subscriber_list)
+        rospy.loginfo('')
```
DSsoto: Comment above also applies here

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -148,8 +160,11 @@ def subscribe_loop(self):
             i = i + 1
             if i % 1000 == 0:
                 rospy.logdebug('still subscribing!')
+        rospy.loginfo("Subscribed to {} of {} topics, will try again every {} seconds".format(self.successful_subscription_count,
```
DSsoto, line 164: Just for your information, you didn't need str.format() here because loginfo is already a formatter. You could've equivalently written:
`rospy.loginfo("Subscribed to %d of %d topics, will try again every %d seconds", self.successful_subscription_count, len(self.subscriber_list), self.resubscribe_period)`
This is true for all the rospy logger functions.

DSsoto (commented):
A couple of minor comments I'd like addressed before merging.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -346,7 +359,7 @@ def start_bagging(self, req):
             # get time index the normal way
             else:
                 bag_index = self.get_time_index(topic, requested_seconds)
-
+            rospy.loginfo('topic: {}, bag_index: {}'.format(topic, bag_index))
```
kev-the-dev: If I'm not familiar with this script's code, I probably won't know what bag_index is. Try to format logging so that the average Joe Smoe would get something useful from it, like the famous line 265.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -285,8 +297,8 @@ def set_bag_directory(self, filename=''):
         # If directory param is not set, default to $HOME/bags/<date>
         default_dir = self.dir
         if default_dir == None:
-            default_dir = os.path.join(os.environ['HOME'], 'bags' ,str(datetime.date.today()))
-
+            default_dir = os.path.join(os.environ['HOME'], 'bags')
+        default_dir = os.path.join(default_dir, str(datetime.date.today()))
```
kev-the-dev: I added this as a hotfix at test day, but it should probably be more customizable. I suggest adding a boolean param "dated_folder" which, if true, will have this behavior of appending the date to the bag dir.

kev-the-dev (changes requested):
Good fixes, but you should double check that your logging makes sense and serves a purpose. Ideally if a node is operating correctly it shouldn't output a whole lot to the screen, and when it does it should be easy to read and understand.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -148,8 +163,11 @@ def subscribe_loop(self):
             i = i + 1
             if i % 1000 == 0:
                 rospy.logdebug('still subscribing!')
+        rospy.loginfo("Subscribed to {} of {} topics, will try again every {} seconds".format(self.successful_subscription_count, 
+            len(self.subscriber_list), self.resubscribe_period))
+        rospy.Timer(rospy.Duration(self.resubscribe_period), self.subscribe)
 
-    def subscribe(self):
+    def subscribe(self, time_info=None):
```
kev-the-dev: You should cancel the timer if all topics are subscribed to so it can stop trying (it should also print a message when this happens)

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -130,7 +145,7 @@ def __getitem__(self, index):
         for topic in self.subscriber_list:
             self.topic_messages[topic] = SliceableDeque(deque())
 
-        rospy.loginfo('topics status: %s', self.subscriber_list)
+        rospy.loginfo('\nInitial subscriber_list: {}'.format(self.get_subscriber_list(False)))
```
kev-the-dev: rospy.loginfo starts a newline for you. You don't need ```\n```

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -285,8 +300,8 @@ def set_bag_directory(self, filename=''):
         # If directory param is not set, default to $HOME/bags/<date>
         default_dir = self.dir
         if default_dir == None:
```
kev-the-dev, line 304: Please add the boolean check to use the dated folder.

kev-the-dev (changes requested):
Couple more small things I noticed.

utils/mil_tools/nodes/online_bagger.py
kingkevlar: Thanks, fixed

utils/mil_tools/nodes/online_bagger.py
kingkevlar: Oops, missed it the first time, now added

kev-the-dev (commented):
I appreciate the work to tidy this up. I think, combined with Tony's bag script it will be really useful for MIL projects and the ROS community. Good work!

kev-the-dev (approved):

utils/mil_tools/launch/online_bagger_example.yaml
```diff
@@ -1,15 +1,26 @@
...
          ["/stereo/right/image_raw",   20], 
          ["/velodyne_points",         300],
-         ["/ooka",                    100]]
+         ["/tf",                    100]]
 
-# comment out if default home/user/online_bagger is desired
-# bag_package_path: ''  
\ No newline at end of file
+# determines how frequently unavailable topics are resubscribed to defaults to 3.0 seconds:
+resubscribe_period: 3.0 # seconds
```
sentree, line 18: This should make things a lot faster.

sentree (approved):
Good stuff, this should be very useful.

PR #54, opened 2017-05-15
SCRIPTS: Add management features to wsmux

sentree:
This implements the functionality requested in uf-mil/SubjuGator#239. It also adds the functionality to bind to the current workspace if none is passed with the `-b` argument and the ability to unbind an SSH client with the `-u` argument.

### DIFF
--- a/scripts/wsmux.sh
+++ b/scripts/wsmux.sh
@@ -30,7 +30,9 @@ _catkin_ws_complete() {
 }
 
 wsmux() {
+	local WORKING_DIRECTORY=$PWD
 	local SELECTION
+	local FILE
 
 	# Get the list of catkin workspaces in the user's home directory
 	_catkin_ws_complete
@@ -130,6 +132,7 @@ wsmux() {
 				echo "-c		--connect		Select the SSH client's workspace"
 				echo "-h		--help			Display the help menu"
 				echo "-l		--list			List the detected workspaces"
+				echo "-p [WORKSPACE]	--pull			Pull the latest changes from upstream"
 				echo "-s		--show			Show the selected workspace"
 				echo "-u		--unbind		Remove the binding for an SSH client"
 				SELECTION="false"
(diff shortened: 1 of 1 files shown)


scripts/wsmux.sh
```diff
@@ -99,30 +105,97 @@ wsmux() {
...
+
+				if [[ "$SELECTION" != "false" ]]; then
+					for FILE in ~/$SELECTION/src/*; do
+
+						# Verify that the directory is a Git repository
+						if [[ -d $FILE/.git ]]; then
+							cd $FILE
+
+							# Verify that the repository's working directory is clean
+							if $(git diff-index --quiet HEAD --); then
```
kev-the-dev, line 185: So is this checking for the exit code? Looks like this command returns exit code 1 for an unclean working directory. So isn't this the opposite of what you want (assuming if runs for positive exit codes)

scripts/wsmux.sh
```diff
@@ -99,30 +105,97 @@ wsmux() {
...
+
+								# Fetch the most recent changes from upstream
+								git fetch upstream
+
+								# Force the creation of a new master branch based on upstream
+								git checkout -B master upstream/master
```
kev-the-dev: Be sure to handle submodules. I think ```git submodule update``` will do it.

kev-the-dev (changes requested):
I ain't no bash boi but I don't think this will have the desired behavior.

scripts/wsmux.sh
sentree: In bash, an exit code of 0 will be taken as `true` for an if statement. So if the working directory is clean, the command will return 0 and the update will proceed.

scripts/wsmux.sh
sentree: Are you sure that won't update them to the latest version of their origin repository?

sentree: This has been added and tested.

kev-the-dev (approved):

DSsoto (approved):

PR #57, opened 2017-05-23
WAYPOINT: add action error to waypoint validation

RustyBamboo:
Reference to uf-mil/SubjuGator#242.

### DIFF
--- a/utils/mil_msgs/action/MoveTo.action
+++ b/utils/mil_msgs/action/MoveTo.action
@@ -8,5 +8,7 @@ float64 linear_tolerance # distance from goal for result to be sent
 float64 angular_tolerance
 ---
 # result
+string error # Returns waypoint error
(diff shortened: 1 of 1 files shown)


utils/mil_msgs/action/MoveTo.action
```diff
@@ -8,5 +8,6 @@ float64 linear_tolerance # distance from goal for result to be sent
 float64 angular_tolerance
 ---
 # result
+int32 error # Returns waypoint error
```
kev-the-dev: This would be better as:
 ```
string error
bool success
```
like other ros services/actions so it is clear to programs using this what the error means.

kev-the-dev (dismissed):

sentree (approved):
This has been open too long for a couple lines of diff.

DSsoto:
Don't be too rash, we probably missed some bugs in there.

PR #58, opened 2017-05-23
FORMAT: auto format C++ code with clang-format

kev-the-dev:
gatta get those green pluses while the getting is good


sentree:
I'm starting to see why the green pluses aren't a reliable metric... Apart from ros_alarms, which @DSsoto has already fixed, the only file that appears to need changes is `perception/mil_vision/include/camera_lidar_transformer.hpp`.

DSsoto (approved):
@ironmig said he was able to compile and run the sim without issues with these changes.

PR #60, opened 2017-05-23
BUG: fix small bugs in mil_vision and mil_ros_tools

kev-the-dev:
Found some bugs running old code. Fixed em' #twolinediff

* in the easy_thresh script, use a different function for BoxPoints if on opencv3

* fix some import nonsense in geometry_helpers

### DIFF
--- a/perception/mil_vision/ros_tools/easy_thresh.py
+++ b/perception/mil_vision/ros_tools/easy_thresh.py
@@ -65,7 +65,7 @@ class Segmenter(object):
 
         self.is_done = False
         rect = cv2.minAreaRect(np.array([np.array(self.corners)]))
-        box = cv2.cv.BoxPoints(rect)
+        box = cv2.boxPoints(rect)
         box = np.int0(box)
         return box
 
--- a/utils/mil_tools/mil_ros_tools/geometry_helpers.py
+++ b/utils/mil_tools/mil_ros_tools/geometry_helpers.py
@@ -1,6 +1,6 @@
 from __future__ import division
 import numpy as np
-from tf.transformations import quaternion_from_euler, euler_from_quaternion, random_quaternion
+import tf.transformations as trans
 from msg_helpers import numpy_quat_pair_to_pose
 from geometry_msgs.msg import Quaternion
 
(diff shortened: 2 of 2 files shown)


utils/mil_tools/mil_ros_tools/geometry_helpers.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import division
 import numpy as np
-from tf.transformations import quaternion_from_euler, euler_from_quaternion, random_quaternion
+from tf.transformations import quaternion_from_euler, euler_from_quaternion, random_quaternion, quaternion_multiply
+from tf.transformations import quaternion_matrix as tf_quaternion_matrix
```
DSsoto: We don't really have these style of imports in our codebase. I think consistency is nice. if you just do `import tf.transformations as trans` then you could just do `trans.quaternion_multiply()`. That way it's clear that all of these share the same namespace.

perception/mil_vision/ros_tools/easy_thresh.py
```diff
@@ -65,7 +65,10 @@ def segment(self):
 
         self.is_done = False
         rect = cv2.minAreaRect(np.array([np.array(self.corners)]))
-        box = cv2.cv.BoxPoints(rect)
+        if cv2.__version__[0] == '2':
+            box = cv2.cv.BoxPoints(rect)
+        else:
+            box = cv2.boxPoints(rect)
```
DSsoto: No need to support multiple versions. The rest of our code is just compatible with Opencv3 (which this doesn't even check for).

DSsoto (changes requested):
I don't really think this is the best way to deal with these, but good on you for trying to clean up old bugs. Awaiting responses.

DSsoto (approved):

DSsoto:
Can you squash the commits? For single commit PRs, requested changes should be ammended to the original commit as per the [code policy](https://github.com/uf-mil/mil_common/wiki/Code-Policy#review-proces).

Also, in the future, consider just appending small impact commits like these to existing PR's so we don't muddy the history too much with unnecessary merge commits.

kev-the-dev:
Squashed 'em. If you don't like merge commits you can always click rebase instead. I think I'd rather small bug fix PRs be submitted than to put these changes in unrelated PRs that might take much longer to get merged.

PR #61, opened 2017-05-25
VISION TOOLS: add colorspace visualizer tool

DSsoto:
This tool allows users to decompose an image into the RGB, HSV, and LAB colorspaces. It is meant to help perception code writers determine in which channel it would easiest to segment out an object of interest using thresholding.

image

### DIFF
--- a/perception/mil_vision/mil_vision_tools/__init__.py
+++ b/perception/mil_vision/mil_vision_tools/__init__.py
@@ -1,0 +1,1 @@
+from rect_finder import RectFinder
--- a/perception/mil_vision/mil_vision_tools/rect_finder.py
+++ b/perception/mil_vision/mil_vision_tools/rect_finder.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python
+import numpy as np
+from geometry_msgs.msg import Polygon
+from mil_ros_tools.msg_helpers import rosmsg_to_numpy, numpy_to_polygon
+import cv2
+
+__author__ = "Kevin Allen"
+
+
+class RectFinder(object):
+    '''
+    Keeps a model of a rectangle in meters, providing utility functions
+    to find 2D and 3D pose estimations of rectangles matching this model
+    found in a computer vision program.
+
+    See github.com/uf-mil/SubjuGator/blob/master/perception/sub8_perception/nodes/path_marker_finder.py
+    for an example of this used in a real time computer vision application.
+    '''
+    def __init__(self, length=1.0, width=1.0):
+        '''
+        Initializes the internal model of the rectangle.
+        Also see RectFinder.from_polygon for creating this from a ROS message
+
+        @param length the measurment of the longer side of the rectangle in meters
+        @param width the measurment of the longer side of the rectangle in meters
+
+        If width > length, the two will be reversed so that length is
+        always the longer side.
+        '''
+        # Ensure length >= width
+        self.width, self.length = tuple(np.sort([length, width]))
+        self.model_3D = np.array([[self.length / 2.0, -self.width / 2.0, 0],
+                                  [self.length / 2.0, self.width / 2.0, 0],
+                                  [-self.length / 2.0, self.width / 2.0, 0],
+                                  [-self.length / 2.0, -self.width / 2.0, 0]], dtype=np.float)
+        # multiplied by 10000 for 0.1mm accuracy in integer image
+        self.model_2D = np.array([[[0, 0]],
+                                  [[self.width * 10000, 0]],
+                                  [[self.width * 10000, self.length * 10000]],
+                                  [[0, self.length * 10000]]], dtype=np.int)
+
+    @classmethod
+    def from_polygon(cls, polygon):
+        '''
+        Creates a RectFinder from a geometry_msgs/Polygon message.
+
+        The length of the Rect becomes the range of x/y (whichever larger)
+
+        The width of the Rect becomes the range of x/y (whichever shorter)
+        '''
+        assert isinstance(polygon, Polygon)
+        arr = rosmsg_to_numpy(polygon.points)
+        # If it contains just one point, treat the x/y and length/width
+        if arr.shape[0] == 1:
+            return cls(arr[:, 0][0], arr[:, 1][0])
+        else:
+            return cls(np.ptp(arr[:, 0]), np.ptp(arr[:, 1]))
+
+    def to_polygon(self):
+        '''
+        Returns a geometry_msgs/Polygon representing the four corners
+        of the rectangle where all z = 0
+        '''
+        return numpy_to_polygon(self.model_3D)
+
+    @staticmethod
+    def sort_corners(rect, debug_image=None):
+        '''
+        Given a contour of 4 points, returns the same 4 points sorted in a known way.
+        Used so that indicies of contour line up to that in model for cv2.solvePnp
+
+        @param debug_image if not None, puts a circle and text of the index in each corner
+
+        p[0] = Top left corner         0--1     1--------2
+        p[1] = Top right corner        |  |  or |        |
+        p[2] = Bottom right corner     |  |     0--------3
+        p[3] = Bottom left corner      3--2
+
+        Credit to David Soto for this implementation.
+        '''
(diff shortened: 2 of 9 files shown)


perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('-c', '--use_cam', help='Use webcam', action='store_true')
+    parser.add_argument('-v', '--use_video', help='Use video', type=str)
+    parser.add_argument('-r', '--use_ros_topic', help='Use ROS topic', type=str)
+    parser.add_argument('-s', '--scale', help='Scale for image processing', type=float, default=0.5)
+    args = parser.parse_args()
+
+    if args.use_cam or args.use_video:
+        cap = None
+        if args.use_cam:
+            cap = cv2.VideoCapture(0)
```
sentree: It may be a good idea to have the device as a parameter for the `--use_cam` argument.

perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
...
+            cap = cv2.VideoCapture(args.use_video)
+
+        while(True):
+            success, rgb_orig = cap.read()
+
+            if not success:
+                print("Unable to get frame")
+                break
+
+            cv2.imshow('Colorspaces - Press any key to exit', viz_colorspaces(rgb_orig, args.scale))
```
sentree: Just a suggestion, but it may be a good idea to have an option to specify the scale of the window in pixels, or to just fit the window to the screen. I can see this getting annoying with images of different sizes.

perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
...
+    hsv = cv2.cvtColor(rgb_small, cv2.COLOR_BGR2HSV)
+    lab = cv2.cvtColor(rgb_small, cv2.COLOR_BGR2LAB)
+
+    viz[h:h * 2, :w] = rgb_small
+    viz[:h, w:w * 2, :], viz[:h, w * 2:w * 3, :], viz[:h, w * 3:w * 4, :] = \
+        map(lambda x: np.repeat(x[:, :, np.newaxis], 3, axis=2), cv2.split(rgb_small))
+    viz[h:h * 2, w:w * 2, :], viz[h:h * 2, w * 2:w * 3, :], viz[h:h * 2, w * 3:w * 4, :] = \
+        map(lambda x: np.repeat(x[:, :, np.newaxis], 3, axis=2), cv2.split(hsv))
+    viz[h * 2:, w:w * 2, :], viz[h * 2:, w * 2:w * 3, :], viz[h * 2:, w * 3:w * 4, :] = \
+        map(lambda x: np.repeat(x[:, :, np.newaxis], 3, axis=2), cv2.split(lab))
```
sentree: This is some compact Python.

sentree (approved):
Great simple tool. I'm surprised this isn't already a ROS utility.

perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python
+from __future__ import division, print_function
```
mattlangford: Ew `print_function`

perception/mil_vision/ros_tools/colorspaces.py
mattlangford: Compact yes, but pretty messy. I'd think something like this would be a better approach (I haven't tested this entire thing):
```python
# Add label to each channel in each image
rgb = np.copy(rgb_small)
cv2.putText(rgb[:, :, 0], 'R', ...)
cv2.putText(rgb[:, :, 1], 'G', ...)
cv2.putText(rgb[:, :, 2], 'B', ...)
cv2.putText(hsv[:, :, 0], 'H', ...)
# etc for all channels

# Construct 3x3 grid of color spaces where each row is RGB, HSV, or LAB respectively
extract_depth = lambda img: img[:, :, 0], img[:, :, 1], img[:, :, 2]
grid = np.vstack((np.hstack(extract_depth(rgb)), np.hstack(extract_depth(hsv)), np.hstack(extract_depth(lab))))

# The grid needs to be 3 channels deep for displaying
grid = np.repeat(grid[:, :, np.newaxis], 3, axis=2)

# build original RGB small image on the side
blank = np.zeros_like(rgb_small)
side_display = np.vstack((blank, rgb_small, blank))

viz = np.hstack((side_display, grid))
```
That way, no weird formula for any positional setting for the text (it's in the same place on each image) and much more readable imo.

And this could be simplified even more I bet.

perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-c', '--use_cam', help='Use webcam', action='store_true')
+    parser.add_argument('-v', '--use_video', help='Use video', type=str)
+    parser.add_argument('-r', '--use_ros_topic', help='Use ROS topic', type=str)
+    parser.add_argument('-s', '--scale', help='Scale for image processing', type=float, default=0.5)
+    args = parser.parse_args()
+
+    if args.use_cam or args.use_video:
+        cap = None
```
mattlangford: Does `cap` need to be `None`'d out?

perception/mil_vision/ros_tools/colorspaces.py
```diff
@@ -0,0 +1,87 @@
...
+
+    elif args.use_ros_topic is not None:
+        key = None
+
+        def disp_image(img):
+            cv2.imshow('Colorspaces - Press any key to exit', viz_colorspaces(img, args.scale))
+            key = cv2.waitKey(10)
+
+        import rospy
+        import mil_ros_tools as mrt
```
mattlangford: It seems like we're doing a lot of abbreviations like this, is there a place we keep a list of standard abbreviations?  I would tend toward not abbreviating especially in trivial cases like this where it's only used once.

mattlangford (changes requested):
I would like the visualization code cleaned up before merging

perception/mil_vision/ros_tools/colorspaces.py
sentree: This is reasonable. It just means there's one less thing to worry about when the lab inevitably moves to Python 3.

mattlangford: I feel like that's not true, the transition will be likely automated right? Like a `flake_8` type thing we run over the whole repo. This just causes confusion in the mean time.

sentree: I guess it would be simple enough to automate for print statements. Honestly, I think we should just make the switch after competition; there aren't many breaking as far as I can tell. We don't want to be on the wrong side of history!

perception/mil_vision/ros_tools/colorspaces.py
DSsoto: Isn't it pretty trivial to just look at the top of the file for the definition of this abbreviation?

I agree about not abbreviating when it's just used once though.

mattlangford: Yeah but if someone was looking through code and there's like 5 different abbreviated things it could be annoying to parse.

perception/mil_vision/ros_tools/colorspaces.py
DSsoto: Sure, back to the old hickory ham print statement.

perception/mil_vision/ros_tools/colorspaces.py
DSsoto: No

perception/mil_vision/ros_tools/colorspaces.py
DSsoto: I'm just trying to set a trend.

kingkevlar:
Oops. I... accidentally closed this trying to push a feature I added... with a force push. I'm so sorry :'(...

@DSsoto  
Would you like me to reopen it? Or would you like to reopen it and stack my commit on top?

PR #55, opened 2017-05-19
CI: Add C++ and Python formatting as build checks

sentree:
This adds a stage to the end of the build process that checks whether or not the new code is formatted according to our style guides. The build will fail if the files are not formatted correctly and a list of the ones that need to be changed for each language will be printed. See the [mil_common wiki](https://github.com/uf-mil/mil_common/wiki) for more information on how to automatically format code.

Have a look at [this Jenkins build](https://ci.mil.ufl.edu/jenkins/blue/organizations/jenkins/uf-mil%2Fmil_common/detail/PR-55/3/pipeline) for this for information on what Python files need to be fixed.

Have a look at [this Jenkins build](https://ci.mil.ufl.edu/jenkins/blue/organizations/jenkins/uf-mil%2Fmil_common/detail/PR-55/2/pipeline) for this for information on what C++ files need to be fixed. @DSsoto, this is all yours if you want it.

Ignore issues in ros_alarms and txros for now. These will be fixed once the formatting pull requests on those repositories are merged and the submodules are updated in this one.

As each formatting pull request is merged to master, the build of this pull request should be run again by the user. Once all of the formatting checks pass, we can merge this. If you disagree with one of the formatting rules, feel free to debate it on this pull request and I will consider removing it.

PR #63, opened 2017-05-26
FORMAT: resolve pep8 errors for python files

kev-the-dev:
* run autopep8 on all python files in mil_common besides ros_alarms
* manually resolve remaining errors outputted by flake8
* I didn't bother fixing manual errors in mil_passive_sonar (there's a lot) because most of it will be replaced when #17 is merged... so not sure what to do bout that (maybe we should just merge #17 in a semi-complete state and improve later)
* How I verified: 
```
python2.7 -m flake8 --max-line-length=120 --exclude=__init__.py,txros,ros_alarms,drivers/mil_passive_sonar .
```


sentree (approved):
I was able to verify the output on this; everything looks good here. We will definitely need to format everything in `drivers/mil_passive_sonar` in #17 and merge that before #55 can be merged, though.

PR #67, opened 2017-06-02
RECT FINDER: try multiple epsilon factors for polygon approx

kev-the-dev:
Cute implementation of @DSsoto's idea to iterate through multiple epsilon values in approximating 4 sided polygon for RectFinder. Should be more robust at finding rectangles in a contour, and be more accurate by starting with only small error tolerances and growing. 

* Note: will break path marker perception in SubjuGator as the API has changed, but I'll append my fix to uf-mil/Subjugator#275 once this is merged.
* Tested by running PathMarker unit test, which uses this. No regressions found.

### DIFF
--- a/perception/mil_vision/mil_vision_tools/rect_finder.py
+++ b/perception/mil_vision/mil_vision_tools/rect_finder.py
@@ -114,7 +114,7 @@ class RectFinder(object):
         '''
         return cv2.matchShapes(contour, self.model_2D, 3, 0.0)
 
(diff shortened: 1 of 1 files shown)


DSsoto (approved):

sentree:
This brings up the interesting fact that we currently have no way of knowing (from the CI, at least) if a change to mil_common will break one of the projects that depend on it. I find that a little worrying, but how do you guys feel about it?

Issue #69, opened 2017-06-07
Make online bagger an action server (hear me out)


kev-the-dev:
I think it would be more appropriate for the triggering of bag writes from online bagger to be through the ROS action API than the ROS service API. 

Why?
* writing large (many gigabyte) bags to disk can take several seconds, which will cause clients (such as alarm callbacks, missions, etc) to freeze for this time
* actions allow a feedback, which will notify the user from the client side which topics are buffered, how many messages, etc instead of having to look through online bagger logs
* actions have built in preempt functionality, so new bag actions can be canceled or superseed current ones
* would make future functionality, like storing new messages even as a bug takes place easier to implement

kingkevlar:
You have been heard

PR #71, opened 2017-06-13
ONLINE BAGGER: convert to action server, add action client cli

kev-the-dev:
* address #69 ;)
* modify bag.sh to be compatible with new system


utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+    parser.add_argument('-d', '--duration', dest='time', type=float, default=0.0,
+                        help='Time in seconds to bag when running as client')
+    parser.add_argument('-t', '--topics', dest='topics', type=str, nargs='+', metavar='topic',
+                        help='List of topics to include in bag when running in client, bag all topics if not set')
+    parser.add_argument('-n', '--name', dest='name', type=str, default='',
+                        help='name of bag to create when running as client')
+    args = parser.parse_args(argv[1:])
+    if args.client:  # Run as actionclient
+        rospy.init_node('online_bagger_client', anonymous=True)
+        if args.topics is None:
```
mattlangford, line 450: You can set a `default` value of arguments in the `add_argument` function.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+            topics = ''
+        else:
+            topics = ''.join(args.topics)
```
mattlangford, line 453: And probably put the type as `list`, though I'm not sure about that one

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+        self.result = None
+        self.last_progress = 0
+
+    def _done_cb(self, status, result):
+        self.result = (status, result)
+
+    def _feedback_cb(self, feedback):
+        iterations = int((feedback.progress - self.last_progress) / 0.01)
+        if iterations >= 1:
+            rospy.set_param('/test', feedback.progress)
```
mattlangford: Remove this?

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -42,11 +42,11 @@ def __init__(self):
         self.get_params()
         self.make_dicts()
 
-        self.bagging_service = rospy.Service('~dump', BaggerCommands,
-                                             self.start_bagging)
+        self._action_server = SimpleActionServer('~bag', BagAction, execute_cb=self.start_bagging, auto_start=False)
```
mattlangford: Since the topic is hard coded down below, you may want to hard code it here too (or vise versa)? You could unknowingly change the namespace and the action client wouldn't be able to connect

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+
+            feedback = BagFeedback()
+            total_messages = 0
+            bag_topics = {}
+            for topic, (time, subscribed) in self.subscriber_list.iteritems():
+                if not subscribed:
+                    continue
+                # Exclude topics that aren't in topics service argument
+                # If topics argument is empty string, include all topics
+                if len(selected_topics) > 0 and topic not in selected_topics:
```
mattlangford, line 350: The first check is redundant

mattlangford (commented):
This is a good idea!

utils/mil_msgs/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ find_package(catkin
 
 add_action_files(FILES
   MoveTo.action
+  Bag.action
```
sentree: This seems a bit generic, it might be a good idea to include "Online" in the name.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -8,21 +8,21 @@
 from collections import deque
 import itertools
 import datetime
-from mil_msgs.srv import BaggerCommands
+from actionlib import SimpleActionServer, SimpleActionClient, TerminalState
+from mil_msgs.msg import BagAction, BagResult, BagFeedback, BagGoal
+import argparse
+from tqdm import tqdm
```
sentree: I knew you would come around!

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
 
+class OnlineBaggerClient(object):
+    '''
+    Wrapper to run an action client connecting to online bagger
+    and triggering a write with the specified topics, time, and
+    filename. Uses tqdm to display a progress bar as the write
+    occurs.
+    '''
+    def __init__(self, name='', topics='', time=0.0):
+        self.client = SimpleActionClient('/online_bagger/bag', BagAction)
```
sentree: Try to stick with the relative topic name, it will reduce the potential for developer error later down the line.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
-                if messages % 100 == 0:  # log every 100th topic, type and time
-                    rospy.logdebug('topic: %s, message type: %s, time: %s',
-                                   topic, type(msgs[1]), type(msgs[0]))
-
-            # empty deque when done writing to bag
-            self.topic_messages[topic].clear()
-
-        self.bag.close()
-        rospy.loginfo('bagging finished!')
+            result.status = 'Bag Request came in while bagging, priority given to prior request'
```
sentree, line 328: This has me wondering how hard dumping separate bags from the queue would be. I don't expect this to be implemented in this pull request, but hear me out:

* When a dump request comes in, it is marked with a `dump_start` and `dump_end` index in the current queue (for the case of one bag at a time, this would just be the beginning and end of the queue)
* When a bag is being dumped, the deletion of old queue entries is halted (but new entries are still being added) and a `deletion_head` is created that moves forward in the queue and marks how many values would have been deleted if no bag was being dumped
* The bag dumped for the initial request above would stop dumping once it reached the `dump_end` marker
* If a second dump request is received while the first one is still running, the `dump_start` index will be set to the current index of the `deletion_head` (that is where the queue would have started if the queue deletion was not paused for the other bag)
* Once a bag finishes dumping, any items at an index before the `deletion_head` and any existing `dump_start` markers is deleted from the queue
* Once there are no bags being dumped, the deletion head can be deleted and the normal rolling queue can be resumed

Basically, this would allow us to dump multiple bags at the same time while still maintaining a rolling queue of messages (for the most part).

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+
+                if req.bag_time == 0:
+                    index = 0
+                else:
+                    index = self.get_time_index(topic, requested_seconds)
+                    total_messages += len(self.topic_messages[topic][index:])
+                    bag_topics[topic] = index
+            if total_messages == 0:
+                result.success = False
+                result.status = 'no messages'
```
sentree, line 363: This is a great debugging message to have.

utils/mil_tools/nodes/online_bagger.py
```diff
@@ -319,78 +313,146 @@ def set_bag_directory(self, filename=''):
...
+            self.bar.update(iterations)
+            self.bar.refresh()
+            self.last_progress = feedback.progress
+
+    def bag(self, timeout=rospy.Duration(0)):
+        self.client.wait_for_server()
+        self.result = None
+        self.last_progress = 0.0
+        self.bar = tqdm(desc='Writing bag', unit=' percent', total=100,
+                        bar_format='{desc} {bar} {percentage:3.0f}%')
```
sentree, line 423: Where does this actually display since it runs on the action server?

sentree (commented):
Great job on this! Hopefully the third time is the charm.

utils/mil_tools/nodes/online_bagger.py
kev-the-dev: I don't think so. The functionality is that if ```topics==''```, include all topics. Otherwise include only the topics in this string

utils/mil_tools/nodes/online_bagger.py
kev-the-dev: I think I prefer the [nargs option](https://stackoverflow.com/questions/15753701/argparse-option-for-passing-a-list-as-option), so the user doesn't have to put the topics in quotes

utils/mil_tools/nodes/online_bagger.py
kev-the-dev: Looks where this code is. This is only run when online bagger is run with ```-c```, which runs an action client.

utils/mil_tools/nodes/online_bagger.py
kev-the-dev: Good catch

utils/mil_tools/nodes/online_bagger.py
kev-the-dev: Yeah that would be pretty useful (especially as a ROS community tool). Onward with online bagger development!

utils/mil_msgs/CMakeLists.txt
kev-the-dev: renamed

sentree (approved):

PR #72, opened 2017-06-14
ONLINE BAGGER: fix bugs from test day

kev-the-dev:
* progress bar now always goes to 100%
* fix indention error introduced during formating which causes no messages to be found if bag_time is not specified
* ```bag -o``` now includes ```$BAG_ALWAYS```

### DIFF
--- a/utils/mil_tools/nodes/online_bagger.py
+++ b/utils/mil_tools/nodes/online_bagger.py
@@ -409,7 +409,7 @@ class OnlineBaggerClient(object):
         self.result = (status, result)
 
     def _feedback_cb(self, feedback):
-        percentage = int(100.0*feedback.progress)
+        percentage = int(100.0 * feedback.progress)
         if percentage - self.total_it >= 1:
             self.bar.update(percentage - self.total_it)
             self.bar.refresh()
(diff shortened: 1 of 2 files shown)


utils/mil_tools/nodes/online_bagger.py
```diff
@@ -403,28 +403,29 @@ def __init__(self, name='', topics='', time=0.0):
...
 
     def _done_cb(self, status, result):
         self.result = (status, result)
 
     def _feedback_cb(self, feedback):
-        iterations = int((feedback.progress - self.last_progress) / 0.01)
-        if iterations >= 1:
-            self.bar.update(iterations)
+        percentage = int(100.0 * feedback.progress)
+        if percentage - self.total_it >= 1:
```
DSsoto, line 413: I've always hated software that did this kind of trickery, but whatever...

DSsoto (approved):

kev-the-dev:
I still don't have merge access by the way. Let's merge this one

sentree (approved):
Looks like you found all of the ghosts Bob Ross!

DSsoto:
You have the exact same permissions on here as you do on the sub repo. Not really sure what to change. We can look at it together if you want.

PR #76, opened 2017-07-06
SCRIPTS: make cm alias leave a free core

DSsoto:
* Detects the number of cores on your system and sets the value of the
  -j* flag to use one less than the number of available cores


DSsoto:
Tested on another machine. The j flag dictates the number of background jobs make will use but does not necessarily mean that number of cores will be used.

PR #80, opened 2017-07-15
MODULE: Add pointgrey_camera_driver as a submodule

sentree:
This is the second of two pull requests that address uf-mil/installer#18. It adds the `pointgrey_camera_driver` as a submodule so that it will be built with our software stack and we can control which version we use. I have also fixed the pull option in `wsmux` so that it handles submodules that have not already been initialized by the user.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ dockerNode(image: 'uf-mil:mil_common') {
 	}
 	stage("Format") {
 		sh '''
-			if [[ ! -z "$(python2.7 -m flake8 --ignore E731 --max-line-length=120 --exclude=__init__.py,drivers/mil_passive_sonar .)" ]]; then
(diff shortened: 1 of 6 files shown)


kev-the-dev (approved):
Good work. Should wait till rebuilt CI docker image succeeds before merging

Jenkinsfile
```diff
@@ -24,15 +24,15 @@ dockerNode(image: 'uf-mil:mil_common') {
 	}
 	stage("Format") {
 		sh '''
-			if [[ ! -z "$(python2.7 -m flake8 --ignore E731 --max-line-length=120 --exclude=__init__.py,drivers/mil_passive_sonar .)" ]]; then
+			if [[ ! -z "$(python2.7 -m flake8 --ignore E731 --max-line-length=120 --exclude=__init__.py,drivers/pointgrey_camera_driver,drivers/mil_passive_sonar .)" ]]; then
```
kev-the-dev, line 27: Update the mcfmt alias too

kev-the-dev (changes requested):

kev-the-dev (approved):

PR #93, opened 2017-12-09
program to find all files written in python

brooksturtle:
fixed as per Kevin's request. Works across any directory and is all within one python program.
pythonfinderex1
pythonfinderex2

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,9 +24,13 @@ dockerNode(image: 'uf-mil:mil_common') {
 	}
 	stage("Format") {
 		sh '''
-			if [[ ! -z "$(python2.7 -m flake8 --ignore E731 --max-line-length=120 --exclude=__init__.py,drivers/pointgrey_camera_driver,drivers/velodyne,drivers/mil_passive_sonar .)" ]]; then
-				echo "The preceding Python following files are not formatted correctly"
-				exit 1
+			echo "Checking Python for Pep8"
+			OUTPUT=$(rosrun mil_tools pythonFinder.py $HOME/mil_ws/src)
+			echo "${OUTPUT}"
+			if [ ! -z "$OUTPUT" ]
+			then
+  				echo "Errors in Python Pep8 formatting"
+  				exit 1
 			fi
 		'''
 		sh '''
--- a/utils/mil_tools/scripts/ReadMe.txt
+++ b/utils/mil_tools/scripts/ReadMe.txt
@@ -1,0 +1,9 @@
+sProgramFinder:
+Finds all files in a directory with either .py endings or #!/usr/bin/env in first line.
+Then checks them for pep8 standards
+
+Type in directory to search from as a second argument
+
+This is how I run it:
+
+rosrun mil_tools pythonFinder.py $HOME/mil_ws/src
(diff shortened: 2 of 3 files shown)


utils/mil_tools/pythonProgramFinder/pythonPrograms.txt
```diff
@@ -0,0 +1,647 @@
+/home/lewis/mil_ws/devel/setup.zsh
```
kev-the-dev: This shouldn't be in the PR, it's just the program output.

utils/mil_tools/pythonProgramFinder/pythonFinder.py
```diff
@@ -0,0 +1,59 @@
+import os
```
kev-the-dev: This script should be executable using ```rosrun mil_tools pythonFinder.py```. For this it should:
1) be given execute permissions ```chmod +x pythonFinder.py```
1) be moved to ```mil_tools/scripts```
1) Start with the magic first line ```#!/usr/bin/env python```

Note that we currently don't use python 3 in our projects. The program should be able to run with python 2 (almost the same)

utils/mil_tools/pythonProgramFinder/pythonFinder.py
```diff
@@ -0,0 +1,59 @@
...
+import re
+
+# Joseph Brooks
+# Checks NaviGator for all python files
+print("Python Finder")
+print("Type directory path you want to search from")
+directory = input(": ")
```
kev-the-dev: This program should take arguments not stdin to pick the folder. It should also default to the current directory if not specified.  Use ```sys.argv```

utils/mil_tools/pythonProgramFinder/pythonFinder.py
```diff
@@ -0,0 +1,59 @@
...
+        programName = os.path.join(root, name)
+        allFileList.append(programName)
+
+
+pythonList = []
+
+# attempts to open file as text and checks #!/usr/bin/env
+for program in allFileList:
+    try:
+        for line in open(program):
```
kev-the-dev: You only have to check the first line. Reading the whole file is going to make this MUCH slower.

utils/mil_tools/pythonProgramFinder/pythonFinder.py
```diff
@@ -0,0 +1,59 @@
...
+# checks for ending .py
+for program in allFileList:
+    if '.py' in program:
+        pythonList.append(program)
+
+# checking for duplicates
+finalList = []
+
+for program in pythonList:
+    # do not add if duplicate
```
kev-the-dev: Python has built in [set functionality](https://docs.python.org/2/library/sets.html) for lists that cannot contain duplicates.

utils/mil_tools/pythonProgramFinder/pythonFinder.py
```diff
@@ -0,0 +1,59 @@
...
+        finalList.append(program)
+
+file = open("pythonPrograms.txt", "w")
+
+counter = 0
+# writing to file
+for program in finalList:
+    file.write(program + "\n")
+    counter = counter + 1
+file.close()
```
kev-the-dev: Remember this is used to pass to flake8 during our [CI process](https://github.com/brooksturtle/mil_common/blob/master/Jenkinsfile). You need to figure out how to pass this to flake8.

kev-the-dev (changes requested):
Good start. Please see comments. The final PR should update the Jenkinsfile to use this program's list of python files.

PR #101, opened 2018-02-10
TOOLS, CI: only run tests in specified directory

kev-the-dev:
Adds a script to only run catkin tests in a certain directory (all packages in that directory), for use in MIL CI so we don't get failing builds for submodules and other projects.


Grymessa (approved):
Looks good, only error I could find was a slight misspelling on line 11 of a comment line so no big deal. Tested it and it built Navigator for me, bounced off incorrect repo names.

RustyBamboo (approved):

PR #104, opened 2018-03-18
Labelbox/ROS interface tooling

kev-the-dev:
* create extract_bag_images script, which dumps images from bags into files according to a yaml configuration. Intended to be used to generate datasets for projects on labelbox.io
* Create ImageProc, a python implementation of image proc which is used in extract_bag_images so it can dump bayered/unrectified images by first debayering/rectifying them
* Create labelbox_parser python class, which will read a json file of annotations exported from labelbox and call a user specified callback with each label and the corresponding image
* Add several utility functions to cv_tools


perception/mil_vision/mil_vision_tools/__init__.py
```diff
@@ -1,3 +1,5 @@
 from shape_finder import RectFinder, EllipseFinder, CircleFinder
-from cv_tools import Threshold, auto_canny
+from cv_tools import *
```
Grymessa, line 2: Going for the blanket import? Bold strategy, Cotton. Lets see if it pays off for him.

perception/mil_vision/mil_vision_tools/labelbox_parser.py
```diff
@@ -0,0 +1,98 @@
...
+
+class LabelBoxParser(object):
+    '''
+    Class to read an exported labelbox annotation file (must be JSON formatted)
+    and call a user specified callback function with this parsed annotation along with the
+    image it annotates.
+    '''
+    def __init__(self, labelfile, image_dir='.'):
+        '''
+        @param labelfile: name of json file containing labelbox annotations
```
Grymessa, line 16: Do we have a labelbox annotation file? I imagine this really wouldn't change from data set to data set unless we plan on segmenting them by tasks (thus limiting the possible annotations). At any rate it might be a good idea to get a default json file for annotations. At some point. Not critical.

Grymessa (approved):
Great tool and thanks for writing it just in time. Using this tonight to get some labelled buoy images and exporting them to pascal for the SSD Network.

Grymessa:
Final note- successfully used this to extract images from bags.

perception/mil_vision/mil_vision_tools/labelbox_parser.py
kev-the-dev: I think you're confused about what labelfile is. This is the json file containing the exported labels from LabelBox. LabelBox also requires a template json file defining what objects can be labeled, but this tool doesn't need to know that.

Grymessa: Yeah I should have updated this after I used LabelBox sorry it contains the info for the labelled images. But I was confused for sure, thought it was for the labelling itself. But that is done in the LabelBox software and is dependent on the mode you choose. Interestingly, you can extract the images from the json afterwards too, separating labels and pngs. Good stuff.

PR #109, opened 2018-03-23
TASKS: cancel goal if task_client closed

### LINKED ISSUE #108: TASKS: cancel task if runtask is clossed
If user Control-C's a runtask command, it should catch this and cancel the task

kev-the-dev:
#108. So now if you messed up on a move command, just control+C that terminal and it will cancel the command


kev-the-dev:
fixes uf-mil/Navigator#319

Gungo (approved):
looking even better

PR #112, opened 2018-04-05
TASKS: print error if cleanup fails

kev-the-dev:


Grymessa (approved):
Better error tracing is always appreciated.

PR #122, opened 2018-05-08
ROS_DOCKER: Add Dockerfile and python script for managing ROS docker

RustyBamboo:
If you can run Docker with a base linux system, then you can run ROS easily! 
Kinda WIP

Dockerfile:
- Install a base ubuntu system
- Creates a mil account
- wget install script, which is to be run by user once starting a container

Python Script (ros_docker_controller)
- Detects no images and automatically builds ROS Image
- ncurses interface to start a new container from ROS Image
- check if existing (even stopped) ROS container is avaliable and loads it
- ncurses interface to open a terminal instance or quit

building

image_selection

manage_container


RustyBamboo:
Moved to uf-mil/installer#31

PR #124, opened 2018-05-31
script to check if teledyne and gumstix are connected

marcgabe15:
is the last line right to check for filesystem usage?

### DIFF
--- a/scripts/100-ping-check
+++ b/scripts/100-ping-check
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+ping_checker() {
+	HOST="$1"
+	HOST_PING="$(ping -c 2 $HOST 2>&1 | grep '% packet' | awk -F'[%]' '{print $1}' | awk -F'[ ]' '{print $NF}')"
+	if [ ! -z "$HOST_PING" ]; then
+		if [ $HOST_PING -lt 25 ]; then
+			echo "$HOST......OKAY"
+		else
+			echo "$HOST......NOT OKAY"
+		fi
+	fi
+}
+ping_checker mil-sub-sub8.ad.mil.ufl.edu
+ping_checker mil-com-teledyne-p900.ad.mil.ufl.edu
+btrfs filesystem usage
--- a/scripts/999-ping-check
+++ b/scripts/999-ping-check
@@ -1,18 +1,21 @@
 #!/bin/bash
 
 ping_checker() {
+	GREEN='\033[32m'
+	RED='\033[31m'
+	NC='\033[0m'
 	HOST="$1"
 	HOST_PING="$(ping -c 2 $HOST 2>&1 | grep '% packet' | awk -F'[%]' '{print $1}' | awk -F'[ ]' '{print $NF}')"
 	if [ ! -z "$HOST_PING" ]; then
 		if [ $HOST_PING -lt 25 ]; then
-			echo "$HOST......OKAY"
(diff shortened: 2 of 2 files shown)


RustyBamboo:
Rename this to like 999-ping-check so that it appears last

scripts/100-ping-check
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+ping_checker() {
+	HOST="$1"
+	HOST_PING="$(ping -c 2 $HOST 2>&1 | grep '% packet' | awk -F'[%]' '{print $1}' | awk -F'[ ]' '{print $NF}')"
```
RustyBamboo: when the host is not available, it takes this over 5 seconds to run, which is a little inconvenient. Any way to reduce this time?

scripts/100-ping-check
```diff
@@ -0,0 +1,16 @@
...
+	if [ ! -z "$HOST_PING" ]; then
+		if [ $HOST_PING -lt 25 ]; then
+			echo "$HOST......OKAY"
+		else
+			echo "$HOST......NOT OKAY"
+		fi
+	fi
+}
+ping_checker mil-sub-sub8.ad.mil.ufl.edu
```
RustyBamboo: shouldn't this be gumstix IP?

scripts/100-ping-check
```diff
@@ -0,0 +1,16 @@
...
+ping_checker mil-com-teledyne-p900.ad.mil.ufl.edu
+btrfs filesystem usage
```
RustyBamboo: `btrfs filesystem usage /`

RustyBamboo:
Yeah, if the host is not available it takes way to long for the motd to appear

RustyBamboo:
image
But it works

marcgabe15:
ok i added a two second timeout does this work better

RustyBamboo:
Address the above requests, and I think it would convenient to have OKAY and NOT OKAY be in color

scripts/100-ping-check
marcgabe15: i added a timeout of two seconds is that okay?

scripts/100-ping-check
marcgabe15: lol my bad

scripts/100-ping-check
marcgabe15: ill fix that too

scripts/100-ping-check
RustyBamboo: :+1:

RustyBamboo:
Okay cool, looks like it works although still not as fast as I'd like it to be.
Last problem is to reformat the commit messages to match our standard of `TOPIC: present tense description`, which can be done by doing interactive rebase `git rebase -i HEAD~2` and then editing the message

scripts/999-ping-check
```diff
@@ -0,0 +1,21 @@
...
+		else
+			echo -e "$HOST......${RED}NOT OKAY${NC}"
+		fi
+	else 
+		echo -e "$HOST...... ${RED} NOT FOUND${NC}"
+	fi
+}
+ping_checker mil-sub-gumstix.ad.mil.ufl.edu
+ping_checker mil-com-teledyne-p900.ad.mil.ufl.edu
+btrfs filesystem usage /
```
RustyBamboo: `btrfs fi show` seems to be a better command to run as it's more straight to the point of file system usage

RustyBamboo (approved):

PR #125, opened 2018-06-06
OBJECT: add msg and srv files

RustyBamboo:


utils/mil_msgs/msg/PerceptionObject.msg
```diff
@@ -0,0 +1,29 @@
+# Time stamp and frame of the object. All objects are in ENU frame.
```
Grymessa, line 1: Is this what real commenting looks like? Its beautiful....

Grymessa (approved):
Good stuff. Stone cold seal of approval.

PR #134, opened 2018-09-25
mil_blueview_driver: still make stub package even if BVT SDK not present to allow other packages (which only depend on message types) to build

forrestv:


kev-the-dev (approved):
Thanks for this, been an annoying issue for a while

PR #139, opened 2018-10-14
SIMULATION: add gazebo plugin to publish heading between two models

kev-the-dev:
Add gazebo plugin to publish a vector between two models. For example, this can be used to simulate the processing of the RobotX pinger, publishing the heading from the hydrophones and the robot


Grymessa (dismissed):
This is included in PR #140 . Was that intentional?

kev-the-dev:
140 depends on 139, hence the "depends" description in the PR comment (also seen in waffle.io). When PRs depend on another, the dependencies must be merged first

Grymessa (approved):
Useful plugin that assists in simulating the pinger.

PR #140, opened 2018-10-14
SIMULATION: add script to set gazebo rtf

kev-the-dev:
depends on #139 
(for boostraping mil_gazebo package)

Adds script to set the real time factor of gazebo, useful for speeding up the simulation / increasing accuracy).


Grymessa (approved):
Good improvements to sim overall.

PR #145, opened 2018-10-28
VISION: add unified computer vision node and messages

### LINKED ISSUE #137: Write unified computer vision interface
Write tooling/library/messages for nodes detecting objects in the 2D frame of a camera, or 3D poses for systems using PnP / stereo.

This system will be similar to the [vision proxies system on NaviGator](https://github.com/uf-mil/NaviGator/blob/master/mission_control/navigator_missions/navigator_missions/navigator.py#L287).

Ideas:
* [ ] A single message shared by vision nodes for reporting what was seen, frame, confidence, additional attributes, etc
* [ ] A base python "server" class for subscribing to a camera, finding objects, and reporting them via this message
* [ ] A python "client" class for missions / visualization tools which connect to vision servers and keeps track of objects / can wait until something is seen. Should provide a TxRos implementation for use with mil_missions
* [ ] A gazebo plugin to report ground truth objects seen in camera
* [ ] An implementation for the "client" which uses tensorflow

kev-the-dev:
closes #137 

This PR intends to unify the several disconnected implementations similar to this, so all MIL computer vision code can share a set of messages, libraries, and tooling. Yay, eliminating duplicate code!

* Adds a set of messages to be used by nodes which identify objects in images (computer vision nodes).
* Adds a python class, VisionNode, which will serve as a base class for computer vision nodes and makes use of these messages. Child classes need to only implement one function to have a fully ros integrated computer vision system that plays nicely with future tools in missions / visualization / testing

### DIFF
--- a/perception/mil_vision/mil_vision_tools/__init__.py
+++ b/perception/mil_vision/mil_vision_tools/__init__.py
@@ -4,3 +4,4 @@ from image_mux import ImageMux
 from image_proc import ImageProc, ImageSet
 from labelbox_parser import LabelBoxParser
 from color_classifier import ContourClassifier, GaussianColorClassifier
+from vision_node import create_object_msg, VisionNode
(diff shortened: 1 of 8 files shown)


Grymessa (approved):
This is an excellent idea and will really streamline development going down the road. I do like the ObjectInImage message. I can see possibly using this to generate some labelled data (that will need to be manually checked but still).

PR #160, opened 2018-11-15
POI: add poi server (GOOD POI)

### LINKED ISSUE #152: Add Point of Interest Node
Add a node to replace the "fake objects" from navigator_lidar_oa. This node will allow humans/programs to place named markers in the global frame marking points of interest. This allows us to program in "priors" on where various challenges are, so the autonomy can explore these regions.

Some features to include:
* Service to spawn POI and set its position
* Service/Topic to get POI
* Rviz interactive markers for moving POI
* Save POIs to a file and load them back at a future time

kev-the-dev:
closes #152

### DIFF
--- a/scripts/bash_aliases.sh
+++ b/scripts/bash_aliases.sh
@@ -73,6 +73,7 @@ alias listtasks="rosrun mil_tasks task_client list"
 # POI
 poisave()
 {
+  rosservice call /poi_server/save_to_param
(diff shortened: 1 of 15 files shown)


kev-the-dev:
Really hate to merge something with this much code without review, but after 14 days this needs to get in so the dependent code can be developed

PR #186, opened 2019-01-31
Add line to CMakeLists.txt to bypass error-throwing

thedpws:


thedpws:
Error about mil_blueview_driver was being thrown on catkin_make ..though the error was caused when my mil_common was outdated. Updating mil_common may have also fixed this error.

PR #198, opened 2019-05-10
DRIVERS: add actuator board CAN handle

Burtt:


drivers/mil_usb_to_can/mil_usb_to_can/device.py
```diff
@@ -35,6 +37,74 @@ def send_data(self, data, can_id=0):
...
+        message = self.IDENTIFIER_BYTE + chr(req.actuator_address) + chr(1)
+        if req.actuator_on:
+            message += chr(1)
+        else:
+            message += chr(0)
+        self.send_data(message)
+        return reduce((lambda a, b: a * pow(2, 8) + b), [ord(char) for char in message])
+
+    def on_get_req(self, req):
+        if not isinstance(req.actuator_address, int) or req.actuator_address < 0 or req.actuator_address > 0x0B:
```
kev-the-dev, line 81: You don't really need to type check ROS messages, they are always serialized with the correct type

kev-the-dev (commented):
This is really good work! Well documented, makes good use of conventions, etc. As you know, I had to implement my own version of this so it could be tested on schedule so this will not be merged in. I'm sure you learned a lot with this project and I look forward to your future work (move test?)

